PR #33, opened 2022-04-04
Instructions to run celestia-bridge via systemd

yelllowsin:
This PR adds instructions to run celestia-bridge via systemd and also makes the current systemd files production grade by increasing LimitNOFILE to 65536.


yelllowsin:
Will try to reorganize it in a way that is OS agnostic.

PR #45, opened 2022-04-11
doc: Add docker-compose resources

YazzyYaz:
Hi @jbowen93 please review this PR

### DIFF
--- a/docs/nodes/devops-resources.md
+++ b/docs/nodes/devops-resources.md
@@ -1,0 +1,7 @@
+# DevOps Resources
+
+This section will be provide an overview of different available resources for Node Operators looking to test out Celestia architecture.
+
+## Docker Compose
+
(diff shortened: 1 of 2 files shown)


docs/nodes/devops-resources.md
```diff
@@ -0,0 +1,7 @@
+# DevOps Resources
+
+This section will be provide an overview of different available resources for Node Operators looking to test out Celestia architecture.
+
+## Docker Compose
+
+* One resource for playing with docker-compose for Celestia infrastructure is the [ephemeral-cluster](https://github.com/celestiaorg/ephemeral-cluster) repository. This repository provides infra-as-code specifications on running a docker-compose instance of Celestia devnet. It is not recommended to be used as-is for running nodes individually but can be used as a reference specification for anyone looking to modify it for their own purposes.
```
jbowen93, line 7: Can we replace `docker-compose` with `Docker Compose` in the two places it's used here. The `docker-compose` binary uses the older version (v1) by default which doesn't work with the ephemeral-cluster.

jbowen93, line 7: I'm not sure we want to refer to the ephemeral cluster as a "reference specification". 
It's not targeted at node operators, it's a place to do integration testing for development.

YazzyYaz: Check the new updated writing.

YazzyYaz: Modified

YazzyYaz:
@jbowen93 please let me know if it's good to go now

jbowen93 (approved):

PR #73, opened 2022-05-16
Add drawio plugin

YazzyYaz:
Tested locally with an example drawio drawing and it works.

Next steps after merging this is moving the existing diagrams that @0xMako has made into their draw.io equivalent and adding them into the docs. For that, would need @0xMako to share the diagrams with me.


adlerjohn (approved):
utACK

PR #86, opened 2022-05-24
Update overview.md

mindstyle85:
note that the links need to be updated here (im bad at these github links lol)

### DIFF
--- a/docs/nodes/overview.md
+++ b/docs/nodes/overview.md
@@ -1,3 +1,42 @@
+
+## Overview
+
+Devnet demonstrates Celestia’s data availability capabilities by running two individual but connected networks:
+
+
+![Network Overview](diagrams/NetworkOverview.png)
+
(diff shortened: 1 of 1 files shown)


YazzyYaz:
@mindstyle85 why do we need to move explanations of the nodes into overview? I think it's better if each explanation is in it's own page like is currently done.

mindstyle85:
i think there needs to be a general overview page where users can see how the whole structure works, before they dive into different node types


also, checking the docs, the validator and bridge node guides are nearly identical, my idea was that the bridge node guide is solely for the celestia-node bridge node, without the app part, since we want non validators to also run those

YazzyYaz:
@mindstyle85 I can modify bridge-node guide to remove Celestia App. I kept it as optional given your recommendations to make it optional in the past.

I don't want to prepopulate the Overview page with lengthy explainers. They should be best kept for each dedicated page. Overview essentially serves as a table of content.

mindstyle85:
yeah, although since we have both celestia app and celestia bridge installation in the validator guide, i think it makes sense to trim down the sole bridge node guide so that people dont get too confused what is what

also, at some point we will need a better general overview of the network somewhere, so that node operators and developers can read it all from one place (imo)

PR #106, opened 2022-05-25
Update environment.md

mindstyle85:
update go version

### DIFF
--- a/docs/developers/environment.md
+++ b/docs/developers/environment.md
@@ -30,7 +30,7 @@ the necessary binaries for running the bridge node. For Golang specifically,
 it’s needed to be able to compile the Celestia Application.
 
 ```sh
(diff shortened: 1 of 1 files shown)


YazzyYaz:
@renaynay @Wondertan @Bidon15 are we good with bumping go version for the dependency setup?

YazzyYaz:
@renaynay @Wondertan @Bidon15 are we good with bumping go version for the dependency setup?

Wondertan (approved):
We do support `1.17` and haven't fully migrated to `1.18`, though I am also fine with encouraging users to rely on the latest Go's versions, which we will migrate to eventually. Our currently update cycle is not strictly defined, but it is something like: "deprecate the usage of the version x once version x+2 comes out".

PR #109, opened 2022-05-26
Bump celestia-node version

mindstyle85:
this is the output now for using the above command, some people check commit here it seems and use the older one which doesnt have the grpc flag in there yet

### DIFF
--- a/docs/developers/celestia-node.md
+++ b/docs/developers/celestia-node.md
@@ -13,10 +13,7 @@ cd $HOME
 rm -rf celestia-node
 git clone https://github.com/celestiaorg/celestia-node.git
 cd celestia-node/
-NODE_VERSION=$(curl -s \ 
-   https://api.github.com/repos/celestiaorg/celestia-node/releases/latest \ 
-   | jq -r ".tag_name") 
-git checkout tags/$NODE_VERSION -b $NODE_VERSION
+git checkout tags/v0.3.0-rc1
 make install
 ```
 
--- a/docs/developers/celestia-node.md
+++ b/docs/developers/celestia-node.md
@@ -26,7 +26,4 @@ Verify that the binary is working and check the version with `celestia version` 
 $ celestia version
 Semantic version: v0.3.0-rc1
 Commit: 5843c5e102651206fe11c685c163f267b141b103
-Build Date: Thu 26 May 2022 02:32:58 PM CEST
-System version: amd64/linux
-Golang version: go1.18.2
 ```
--- a/docs/developers/celestia-node.md
+++ b/docs/developers/celestia-node.md
@@ -13,6 +13,7 @@ cd $HOME
 rm -rf celestia-node
 git clone https://github.com/celestiaorg/celestia-node.git
 cd celestia-node/
+git checkout tags/v0.3.0-rc1
 make install
 ```
 
@@ -20,9 +21,9 @@ Verify that the binary is working and check the version with `celestia version` 
 
 ```console
 $ celestia version
-Semantic version: v0.3.0-rc1-7-g45f23d8
-Commit: 45f23d8ecab4c3526a7729cd4bc9a0ebe561846e
-Build Date: Thu 26 May 2022 10:41:59 AM CEST
+Semantic version: v0.3.0-rc1
+Commit: 5843c5e102651206fe11c685c163f267b141b103
+Build Date: Thu 26 May 2022 02:32:58 PM CEST
 System version: amd64/linux
-Golang version: go1.18.1
+Golang version: go1.18.2
 ```
--- a/docs/developers/celestia-node.md
+++ b/docs/developers/celestia-node.md
@@ -13,7 +13,10 @@ cd $HOME
 rm -rf celestia-node
 git clone https://github.com/celestiaorg/celestia-node.git
 cd celestia-node/
-git checkout tags/v0.3.0-rc1
+APP_VERSION=$(curl -s \ 
+   https://api.github.com/repos/celestiaorg/celestia-app/releases/latest \ 
+   | jq -r ".tag_name") 
+git checkout tags/$APP_VERSION -b $APP_VERSION
 make install
 ```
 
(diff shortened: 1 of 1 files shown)


Wondertan (commented):
So here, you clearly build it from the main branch instead of tagged release. Pls change it to the tagged v0.3.0-rc1 release instead. It does not have some of the changes in main, but those will be included in the next rc candidate.

mindstyle85:
updated it, please review @Wondertan

docs/developers/celestia-node.md
```diff
@@ -13,16 +13,17 @@ cd $HOME
 rm -rf celestia-node
 git clone https://github.com/celestiaorg/celestia-node.git
 cd celestia-node/
+git checkout tags/v0.3.0-rc1
```
adlerjohn, line 16: It might be worth it to use a variation of https://github.com/celestiaorg/docs/blob/e0f76b2067ce50bc6f29eec07495f92a45274f08/docs/developers/celestia-app.md?plain=1#L18-L20

docs/developers/celestia-node.md
```diff
@@ -13,16 +13,17 @@ cd $HOME
...
 $ celestia version
-Semantic version: v0.2.0
-Commit: 1fcf0c0bb5d5a4e18b51cf12440ce86a84cf7a72
-Build Date: Fri 04 Mar 2022 01:15:07 AM CET
+Semantic version: v0.3.0-rc1
+Commit: 5843c5e102651206fe11c685c163f267b141b103
+Build Date: Thu 26 May 2022 02:32:58 PM CEST
 System version: amd64/linux
-Golang version: go1.17.5
+Golang version: go1.18.2
```
adlerjohn, line 28: This isn't easily maintainable. A CI job should be added that runs the above steps, then runs `celestia version` and compares the output to these lines.

docs/developers/celestia-node.md
mindstyle85: did it

docs/developers/celestia-node.md
mindstyle85: Its true, if this changes a lot it isnt maintainable, but for now i think its still better than nothing while theres a better solution. 

We could also just mention that the user needs to check if his version matches the latest release

docs/developers/celestia-node.md
adlerjohn: You used `app` instead of `node`.

docs/developers/celestia-node.md
adlerjohn: #116

docs/developers/celestia-node.md
```diff
@@ -13,16 +13,20 @@ cd $HOME
 rm -rf celestia-node
 git clone https://github.com/celestiaorg/celestia-node.git
 cd celestia-node/
+APP_VERSION=$(curl -s \ 
+   https://api.github.com/repos/celestiaorg/celestia-app/releases/latest \ 
+   | jq -r ".tag_name") 
+git checkout tags/$APP_VERSION -b $APP_VERSION
```
adlerjohn, line 19:
```suggestion
NODE_VERSION=$(curl -s \ 
   https://api.github.com/repos/celestiaorg/celestia-node/releases/latest \ 
   | jq -r ".tag_name") 
git checkout tags/$NODE_VERSION -b $NODE_VERSION
```

docs/developers/celestia-node.md
mindstyle85: whoops, long day i guess

docs/developers/celestia-node.md
```diff
@@ -13,16 +13,20 @@ cd $HOME
...
 Verify that the binary is working and check the version with `celestia version` command:
 
 ```console
 $ celestia version
-Semantic version: v0.2.0
-Commit: 1fcf0c0bb5d5a4e18b51cf12440ce86a84cf7a72
-Build Date: Fri 04 Mar 2022 01:15:07 AM CET
+Semantic version: v0.3.0-rc1
+Commit: 5843c5e102651206fe11c685c163f267b141b103
+Build Date: Thu 26 May 2022 02:32:58 PM CEST
```
Wondertan, line 29: optional not: It might not be worth mentioning Build Date as it is unique to each build

Wondertan (approved):
Thanks @mindstyle85 and thaks to @adlerjohn for catching things I’ve missed

docs/developers/celestia-node.md
mindstyle85: removed the bottom 3 lines as theyre not really important

mindstyle85:
can we merge this? people are installing v0.2.0 still following the docs

mindstyle85:
@Wondertan @adlerjohn that command for getting release version wont work correctly, it pulls down v0.2.0

adlerjohn:
Yes, because v0.3.1-rc-1 is a pre-release. Once an actual release is tagged then it will work. Pre-releases shouldn't be used anyways.

mindstyle85:
> Yes, because v0.3.1-rc-1 is a pre-release. Once an actual release is tagged then it will work. Pre-releases shouldn't be used anyways.

yes i understand that, but the instructions in docs dont work with 0.2.0 as the grcp flag does not exist there, so @renaynay said people should use the v0.3.0-rc1 version

Wondertan:
> Pre-releases shouldn't be used anyways.
@adlerjohn, why? From my understanding, it is helpful to allow users to play around with pre-release for some time and then make a release. Do you suggest making the full release instead already?

Wondertan:
We can hardcore for now, but surely changing it manually all the time is not the good idea, so we might just go with Jonhs suggestion.
> that command for getting release version wont work correctly, it pulls down v0.2.0

adlerjohn:
Realistically, users that are unable to understand how to look up releases on GitHub and pull the latest pre-release shouldn't be using pre-releases; they should be using releases. We should tag an actual release for Mamaki testnet.

mindstyle85:
yes, but one would expect that the instructions provided would work with the version posted, which it doesnt now, so they come back to us on the channel asking for help. I would propose a hardcoded one until an official release is made. Not many of these people know how to properly run nodes anyways, but that doesnt mean we dont need them i think, especially the light nodes

liamsi (approved):

PR #111, opened 2022-05-26
Instructions to use --keyring-backend=test

yelllowsin:
These instructions avoid issues with wallets on current SDK version. Without this, users will get this error:
Error: rpc error: code = NotFound desc = rpc error: code = NotFound desc = account not found: key not found


yelllowsin (commented):
fix line length

YazzyYaz (approved):

PR #112, opened 2022-05-26
Add older network docs

YazzyYaz:


docs/nodes/troubleshoot.md
```diff
@@ -0,0 +1,44 @@
...
+
+```sh
+celestia-appd keys add --recover <keyname>
+# or 
+celestia-appd keys add --recover <keyname> --keyring-backend <other_format>
+``` 
+
+followed by a prompt to enter a bip39 mnemonic, which is the mnemonic that was created as part of `1_create_key.sh` script in the first step.
+
+You'll also be asked for a passphrase which is an input you have to define.
```
Bidon15, line 30: I believe we are not using this script for mamaki, so removing this is ideal 
```suggestion

```

YazzyYaz:
After further discussions, we are closing this PR as all this stuff is outdated.

Issue #122, opened 2022-05-28
PFD curl commands cause errors in bash shell


iofq:
https://docs.celestia.org/developers/node-tutorial#node-api-calls

Working through the above steps:

* `curl -X GET http://127.0.0.1:26658/balance/` returns a 404, the working endpoint for me is `curl -X GET http://127.0.0.1:26658/balance`

* The example PFD curl command gives me the below error. This is fixed by removing the slashes within the curl command.
    * `invalid character '\\\\' looking for beginning of object key string`

Wondering if others are hitting these issues, or if the second issue is just something weird with my shell.

PR #125, opened 2022-05-30
Update node-tutorial.md

YazzyYaz:
Resolves https://github.com/celestiaorg/docs/issues/122


adlerjohn (approved):
utACK

adlerjohn:
Is the second half of #122 not relevant?

>The example PFD curl command gives me the below error. This is fixed by removing the slashes within the curl command.

>    invalid character '\\\\' looking for beginning of object key string

YazzyYaz:
@adlerjohn fixed with the second commit

adlerjohn (approved):

PR #130, opened 2022-05-31
More Link Fixes

YazzyYaz:
* Fixes based on this doc https://docusaurus.io/docs/markdown-features/links

### DIFF
--- a/docs/developers/celestia-app.md
+++ b/docs/developers/celestia-app.md
@@ -3,7 +3,7 @@
 
 This tutorial will guide you through building Celestia App. This
 tutorial presumes you completed the steps in setting up your
-own environment [here](../developers/environment).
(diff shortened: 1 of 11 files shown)


adlerjohn (commented):
1. For links in the current directory, can you use `./filename.md` instead of `filename.md`? It's more explicit.
2. For links in other directories, can you use the relative rather than absolute path? Otherwise it doesn't work on GitHub. For example the link in `docs/developers/node-api.md` is: https://github.com/celestiaorg/docs/blob/7dd0eb5e611e600e226ccd217df6aa91c3f60895/nodes/validator-node.md#deploy-the-celestia-node

YazzyYaz:
@adlerjohn check now if you can

PR #131, opened 2022-05-31
Update light-node.md

mindstyle85:


docs/nodes/light-node.md
```diff
@@ -69,6 +69,14 @@ celestia light start --core.grpc <ip>:9090
 
 If you need a list of RPC endpoints to connect to, you can check from the list [here](../nodes/mamaki-testnet#rpc-endpoints)
 
+Once you start the Light Node, a wallet key will be generated for you.
+You will need to fund that address with some tokens to pay for PayForData transactions.
+You can find the address with:
+
+```sh
+$HOME/celestia-node/cel-key list --home /home/pops/.celestia-light/keys/ --keyring-backend=test
```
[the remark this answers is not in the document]
mindstyle85: ah, forgot to change it, thats from our server

i can make it without the path, so just ./cel-key

mindstyle85:
> The address also gets logged out to you on start.

yeah, most wont see that though, so i think its best we show them how they can check

renaynay:
Would be nice to mention both that it's immediately logged out on start, but also the instructions on how to check it via `cel-key`

mindstyle85:
ok added it, lmk if its ok @adlerjohn

docs/nodes/light-node.md
```diff
@@ -69,6 +69,17 @@ celestia light start --core.grpc <ip>:9090
...
 
+Once you start the Light Node, a wallet key will be generated for you.
+You will need to fund that address with some tokens to pay for PayForData transactions.
+You can find the address by:
+
+1. Checking the logs output
+2. Using the following command:
+
+```sh
+./cel-key list --node.type light --keyring-backend test
```
adlerjohn, line 80: The command `./cel-key` doesn't exist anywhere from the previous steps.

adlerjohn (changes requested):
Please pull from upstream, as it contains a fix to link format.

docs/nodes/light-node.md
mindstyle85: right, the whole guide, and for the other 2 nodes too, is missing that and/or the link to the keys page, should we maybe link the keys page first then? cc @YazzyYaz

mindstyle85:
closing this as unsure how to pull changes in via web interface, will open a new PR, i think its easier for me

PR #132, opened 2022-05-31
Update bridge-node.md

mindstyle85:


adlerjohn:
These are identical changes to #131, but in a different file. For ease of review, please do it all in #131

Issue #63, opened 2022-05-12
Integrate Search Functionality in the Docs


YazzyYaz:
Integration search functionality in the docs will make finding things much easier as we grow our docs.

PR #138, opened 2022-06-07
Add Algolia Search Bar

YazzyYaz:
Closes https://github.com/celestiaorg/docs/issues/63


liamsi (commented):
I did not test this. Also I do not have any context over that search bar vs others. 
Change looks good though.

YazzyYaz:
@liamsi it's based on Algolia being added to Docusaurus which has native support and they gave us a free API key because we are open-source so that it can index our website.

https://docusaurus.io/docs/search#using-algolia-docsearch

liamsi (approved):

Issue #161, opened 2022-07-18
Sentence-case all titles for consistency


rootulp:
https://github.com/celestiaorg/docs/blob/main/docs/developers/celestia-app-commands.md does this. https://github.com/celestiaorg/docs/blob/main/docs/developers/node-tutorial.md does not.

Issue #160, opened 2022-07-18
Consistent casing for light node, bridge node, full node


rootulp:
I don't think these should be capitalized when used in paragraphs. Ex:
1. https://github.com/celestiaorg/docs/blob/main/docs/nodes/systemd.md?plain=1#L205=

rootulp:
This issue isn't resolved yet. Ex. https://github.com/celestiaorg/docs/blob/dfebcb60b2623def5fe9b30e16cad11953c600e3/docs/nodes/systemd.md?plain=1#L205=

YazzyYaz:
@rootulp any updates on this issue? Are you going to make a PR?

rootulp:
Is this issue something maintainers want? Given no one else +1'ed this issue or commented, I may be alone in wanting consistent casing.

YazzyYaz:
@rootulp seems that way.

PR #163, opened 2022-07-18
Sentence case titles

rootulp:
I preserved capitalization for H1s where the noun was the sidebar name (ex. `Full Storage Node`, `Bridge Node`, etc.) and sentence cased all other titles (H2+).

Resolves https://github.com/celestiaorg/docs/issues/161

### DIFF
--- a/docs/nodes/bridge-node.md
+++ b/docs/nodes/bridge-node.md
@@ -1,11 +1,11 @@
-# Setting Up A Celestia Bridge Node
+# Setting up a Celestia Bridge Node
 
 This tutorial will go over the steps to setting up your Celestia Bridge node.
 
 Bridge nodes connect the data availability layer and the consensus layer
 while also having the option of becoming a validator.
 
-## Overview of Bridge Nodes
+## Overview of bridge nodes
 
 A Celestia bridge node has the following properties:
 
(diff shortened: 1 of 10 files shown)


YazzyYaz:
Hey @rootulp I'll get back to this next week, my general thinking though is that title case are fine in sub-headings as they're still titles. 

What is the thinking around lower-casing subheadings below H1? Is that part of a style guide you can link me to?

rootulp:
No rush on reviewing this, next week is fine. We should align on whether we want https://github.com/celestiaorg/docs/issues/161

> What is the thinking around lower-casing subheadings below H1? Is that part of a style guide you can link me to?

It's a style thing. https://design-system.service.gov.uk/styles/typography/#headings states:

> Write all headings in sentence case.

but I'm sure we can find style guides that state the opposite. This PR was motivated by me noticing that the docs contain a lot of capitalization (in titles and paragraphs) which led to my confusion on which words were proper nouns. We probably don't _need_ sentence-cased titles if we do https://github.com/celestiaorg/docs/issues/160 along with downcasing words in paragraphs that aren't proper nouns.

YazzyYaz:
I can go either way, although not too sure on the new styling.

@dabit3 any thoughts?

rootulp:
> I.e. are Light / Bridge / Full / etc... nodes a proper noun? Or is it just at type of Celestia node?

I think just a type of node. For example "light client" and "light node" are not capitalized in:
- https://ethereum.org/en/developers/tutorials/run-light-node-geth/
- https://geth.ethereum.org/docs/interface/les

dabit3:
@rootulp makes sense, looks great to me!

adlerjohn (approved):

rootulp:
I have a lot of other fixes I'd like to make to these docs queued up after this merges

Issue #48, opened 2022-04-14
Explain Why Validators Need to Run A Bridge Node


YazzyYaz:
Explain the need for validators to also need a bridge node.

YazzyYaz:
cc @renaynay @Wondertan @Bidon15

renaynay:
@YazzyYaz in what format would you like for us to do this? Would you like an explanation here that you will turn into docs or would you like for us to write the doc?

YazzyYaz:
We would need an explanation, here is fine. I imagine we need to add it both to the tutorial guide on running a node and maybe in the overview section.

I can help you folks work on one if we want to draft something in Google Docs first.

PR #167, opened 2022-07-27
Added sentence to give more info about bridge node #48

dabit3:


docs/nodes/bridge-node.md
```diff
@@ -5,6 +5,9 @@ This tutorial will go over the steps to setting up your Celestia Bridge node.
 Bridge nodes connect the data availability layer and the consensus layer
 while also having the option of becoming a validator.
 
+Validators do not have to run bridge nodes, but they should because it relays
+blocks to the DA network.
```
rootulp, line 9: I'm confused because 
https://github.com/celestiaorg/docs/blob/542d0ca2b696d7dcf8cd62e3b7c9811ddac31564/docs/nodes/overview.md?plain=1#L9-L10 

suggests that validators do need to run bridge nodes 🤷‍♂️

rootulp: If validators don't need to run bridge nodes, then perhaps a better line in the overview page would be:

```
* [Validator node](./validator-node.md): This type of node participates in consensus by producing and voting on blocks.
```

If so, then happy to merge this PR as-is and create an issue / PR to fix the overview.md page.

rootulp (approved):
Approved with potential follow-up issue

YazzyYaz (approved):

YazzyYaz:
Hi @dabit3 looks like there's merge conflicts, can you resolve those?

docs/nodes/bridge-node.md
dabit3: Hmm, let me cc @renaynay to see if she can maybe provide more insight here.

dabit3: cc @Wondertan any insight on this discussion / description around giving more information about bridge nodes?

renaynay: Validators and bridge nodes are separate node types.

Bridge nodes need a connection to a celestia-core node (that means either validator or non-validating tendermint full node)

We should remove the line mentioned here https://github.com/celestiaorg/docs/pull/167/files#r931620621 from the docs because it is confusing. 

Validators *do not* have to run bridge nodes, but are encouraged to in order to relay blocks to the DA network.

renaynay: The reason we recommend validators to do it is because running a bridge node requires an RPC connection to a validating || non-validating tendermint node, and since validators are already running validators, they can feed their bridge node the RPC address of their own validator and run them in tandem.

PR #171, opened 2022-07-27
fix: type of node on consensus full node steps

rootulp:


dabit3 (approved):

rootulp:
I don't have write access so will defer to @dabit3 to merge. Also open to being granted write access.

Screen Shot 2022-07-27 at 7 21 09 PM

YazzyYaz (approved):

PR #185, opened 2022-08-14
Step 1 : Translations

Gunter038:
Hello @YazzyYaz ! 

I'm adding the first step for the translations.

To begin with, I added 5 languages.  The documents are not translated on this PR. (Except one french code.json). 
Translations will happen on Crowdin. 

I did serveral modifications to translate the frontpage. 

To translate the Sidebar : 

I added a "sidebar_label". The labels are the same as the english sidebar's, it won't change anything. 

On  crowdin we will have : 

image

It enables us to keep only one sidebar, and just translate the labels. 

Once the PR is accepted, we're able to directly add translations. Crowdin creates a PR on my repo. 

On Step 2, we'll add translations, and versioning.


YazzyYaz:
Hi @Gunter038 I have tested out switching languages and it shows as "Page Not Found"

Screen Shot 2022-08-23 at 4 35 04 PM

All I did was go to any of the docs pages and then switch languages and it shows like this. Furthermore, if I try another language, it does something weird in the URL

Screen Shot 2022-08-23 at 4 39 26 PM

YazzyYaz:
Whelps, looks like i forgot to run "yarn build" which installs locale

Issue #59, opened 2022-05-12
Document Celestia's Reserved Namespaces


YazzyYaz:
It's important for developers to know what Celestia's Reserved Namespaces are

PR #206, opened 2022-08-31
Add reserved namespace ids

YazzyYaz:
Closes https://github.com/celestiaorg/docs/issues/59


YazzyYaz:
Taken from here: https://github.com/celestiaorg/celestia-specs/blob/master/src/specs/consensus.md#reserved-namespace-ids

dabit3 (approved):

dabit3:
LGTM, I'm thinking because of the use of tables, it might make sense to disable the line length linting rule here.

PR #250, opened 2022-10-05
Translations : Spanish, French, Ukrainian, Chinese

Gunter038:
Hello ! @YazzyYaz ! 

Here is our first batch. 

As soon as you accept the PR, I'll modify the documents to deal with a "sidebar_label" issue (Crowdin sometimes messes it up), and correct few things. 

Translators will also have a review. 

Translations are not at 100% as you added some changes recently, and I had to modifiy the translations to avoid bugs.


YazzyYaz:
Thank you for your contribution Gunter. I am thinking of automating the process of getting the latest crowdin updates as part of a script. We can do it here https://docusaurus.io/docs/i18n/crowdin

PR #252, opened 2022-10-07
Versioning Testnets

YazzyYaz:
Allows selecting which version of the testnet the developers can select.

Current is Arabica
Latest is Mamaki. 

It defaults to Mamaki. We can switch later when Arabica is retested post network upgrade.

This introduces versioned directories where Mamaki is located in. 

In the future, when Mainnet is ready, the docs reflect Mainnet while Arabica and Mamaki are shown as options in the dropdown.


YazzyYaz:
Closing this in favor of the tabs feature that went live in the docs

PR #257, opened 2022-10-12
gm world tutorial

jcstein:
WIP


jcstein:
btw. if there are any edits on this, i'll add them to english version and overwrite i18n versions

PR #258, opened 2022-10-13
doc: increase gas from 60k to 70k in pfd example

Bidon15:
Add code responses link to learn more about statuses


jcstein:
I think this is an issue with Arabica @YazzyYaz

Bidon15:
> I think this is an issue with Arabica @YazzyYaz

We will need to recheck the documentation once we are done with upgrading https://github.com/celestiaorg/devops/issues/45

YazzyYaz (approved):

PR #261, opened 2022-10-19
Add fork-choice and message layout specs

YazzyYaz:


docs/concepts/specifications/message-block-layout.md
```diff
@@ -0,0 +1,119 @@
+# Message Layout
```
rootulp, line 1: [blocking][question] 

I'm concerned this will increase the maintenance burden for future edits to the specs. In other words, modifications may be made to message_block_layout.md in https://github.com/celestiaorg/celestia-app/pull/796/files and not also updated here due to human error.

What is the rationale for copying the specs content instead of linking to the specs?

YazzyYaz: The rationale is to have all specs in the documentation.

I am happy you guys are updating the specs for celestia-app btw! 

Another thing is I want to highlight some of the specs so they're explainers for the testnet program. Having everything in one website is the best imo.

I agree on maintaining two sets of specs isn't ideal, not sure on the best approach.

rootulp: > The rationale is to have all specs in the documentation.

Sorry for being slow but why do we want want this? Is it so that readers don't have to jump from the docs to the specs via a link? Is it because we think the current specs site (and in the future per-repo specs site) is non-discoverable? 

> Another thing is I want to highlight some of the specs so they're explainers for the testnet program. 

I'm on-board with highlighting particular sections of the specs but they should be individually linkable via headings. I'm not convinced we need to duplicate the content because (sorry for repeating myself) the overhead of maintenance will be higher.

> Having everything in one website is the best imo.

If we want this and don't want two sources of truth then we'd have to move the specs content from per-repo based to all in this repo. Alternatives (suggested by @musalbas / @adlerjohn in Slack) are to use Git submodules or a CI copy + paste workflow.

rootulp: > But I need to merge the PR in order to proceed w other incentivized testnet work

Sorry for blocking this PR and thanks for sharing that this is blocking other work. I'm comfortable approving if we do (in order of preference):
1. remove the duplicate content and linking to the specs
2. duplicating this content if we have a follow-up issue to find an automated mechanism to copy content from the source of truth (per-repo specs)

Approving to unblock.

YazzyYaz: Let's do 2. for now while we scope out the best way to do 1.

jcstein (approved):
Lgtm and looking forward to figuring out best workflow to reduce duplication of effort

PR #267, opened 2022-10-22
Update mamaki-testnet.md

mindstyle85:
fixed validator operator addr in example as people literally copy paste it and this one still using the old format

### DIFF
--- a/docs/nodes/mamaki-testnet.md
+++ b/docs/nodes/mamaki-testnet.md
@@ -133,7 +133,7 @@ example you can run:
 
 ```sh
 celestia-appd tx staking delegate \
-    celestiavaloper1q3v5cugc8cdpud87u4zwy0a74uxkk6u43cv6hd 1000000utia \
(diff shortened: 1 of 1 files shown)


docs/nodes/mamaki-testnet.md
```diff
@@ -133,7 +133,7 @@ example you can run:
 
 ```sh
 celestia-appd tx staking delegate \
-    celestiavaloper1q3v5cugc8cdpud87u4zwy0a74uxkk6u43cv6hd 1000000utia \
+    celestiavaloper1q3v5cugc8cdpud87u4zwy0a74uxkk6u4q4gx4p 1000000utia \
```
YazzyYaz, line 136: What's the change you are making? You are just modifying key value from the looks of it

mindstyle85: yes, the one above is from the old binaries and team changed the bech32 format in between, so people get a mismatch error when using this address from the example

YazzyYaz: got it, thank you

YazzyYaz (approved):

PR #269, opened 2022-10-24
Truncate head response

YazzyYaz:
Long JSON output messes with our crowdin integration, saying the string is too long. A truncated output will resolve this issue.


renaynay (approved):
lgtm but maybe include that this is a truncated header response

Issue #273, opened 2022-10-24
Deploy Preview


MSevey:
Since this repo uses github pages for deployment, we might consider using this github deploy preview action.

https://github.com/marketplace/actions/deploy-pr-preview

YazzyYaz:
Thanks for the link, Matt. I made a PR but need to test it https://github.com/celestiaorg/docs/pull/275

PR #275, opened 2022-10-24
Add GH-Preview Workflow

YazzyYaz:
Closes https://github.com/celestiaorg/docs/issues/273


.github/workflows/preview.yml
```diff
@@ -0,0 +1,27 @@
...
+jobs:
+  deploy-preview:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Install and Build
+        run: yarn install --frozen-lockfile
+        run: yarn build
```
MSevey, line 22:
```suggestion
      - name: Install
        run: yarn install --frozen-lockfile
      - name: Build
        run: yarn build
```

YazzyYaz:
Hrm @MSevey it says `Resource Not Accesible By Integration` error after we add this

YazzyYaz:
Looks like this might be the issue https://github.com/orgs/community/discussions/24701

MSevey:
> Looks like this might be the issue https://github.com/orgs/community/discussions/24701

hmm, I don't think so. Apps are different than actions I believe. This action is simply pushing to a branch at the end of the day.

MSevey:
So I think the issue is because the we can't just use the default for this workflow. As noted in the README, we need additional flags.

https://github.com/celestiaorg/docs#deployment

And also, in the deploy workflow we set a user name

https://github.com/celestiaorg/docs/blob/main/.github/workflows/deploy.yml#L33

Looking at the action, it doesn't seem like there is the flag for that unfortunately. Bummer. I'll open an issue with them and see if they can add it. 

We can put this PR in draft in the meantime.

MSevey:
Hmm, actually, I wonder if we could just use the deploy action we use today and edit it for the previews... I might be able to do that.  Let me play around with it

MSevey:
Progress!

MSevey:
So the last thing needed to make this work is updating the base url for the static site. I'm not really sure where that is configured for docusaurus projects. For React projects for example this is configurable with the `PUBLIC_URL` variable. 

Basically by default this is `/` and we want that for normal deploys. 
For previews, we want it to be `/preview/pr-##` which we can set in the github action with a variable. I just don't know where in the code that variable needs to be referenced fully.

MSevey:
Ok, so I believe the reason for the page not found is not this issue https://github.com/rossjrw/pr-preview-action/issues/12

which is a bummer. I'll comment on the issue asking for any update.

MSevey (approved):

PR #287, opened 2022-11-02
update macOS installation instructions for `celestia-node`

jcstein:
On some macOS, users will sometimes encounter an error when running `make install`, the fix is to run `make go-install`
added to:
- [x] `/developers/celestia-node`
- [x] add tabs for macos v. linux
- [x] `/developers/node-tutorial`
- [x] add tabs for macos v. linux
- [x] `/nodes/light-nodes`
- [x] add tabs for macos v. linux

### DIFF
--- a/docs/developers/celestia-node.mdx
+++ b/docs/developers/celestia-node.mdx
@@ -30,6 +30,22 @@ git checkout tags/v0.4.2
 make install
 ```
 
+You may encounter an error similar to:
+
+```sh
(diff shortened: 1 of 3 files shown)


YazzyYaz:
Can't we just have `make go-install` as default installation for MacOS guides?

jcstein:
:yes: one of my first tasks next week 😎

YazzyYaz (approved):

PR #299, opened 2022-11-08
jcs-recipe-book-rollmint

jcstein:
- [x] initial draft of recipe book
- [x] break out into sections
- [x] add graphics


sidebars.js
```diff
@@ -204,7 +204,7 @@ const sidebars = {
       items: [
         {
           type: "category",
-          label: "gm world",
+          label: "☀️ gm world",
```
YazzyYaz, line 207: why are you adding an asterisk here

YazzyYaz (approved):

PR #358, opened 2022-12-08
Missing installation steps

samlaf:
The ignite cli app installs the runs the app for us when we run `ignite chain serve`. However, we are not starting the app via ignite here. So we need to add steps to build gmd before the `init.sh` script can call it.

## Checklist


samlaf:
Oh also building results in this error
```
cmd/root.go:306:65: undefined: server.FlagDisableIAVLFastNode
```
because celestia renamed this flag `FlagIAVLFastNode` instead... not sure why. So we need to go change that one reference name in root.go.
I'll let you guys either update this doc, or update the celestia variable name.

jcstein:
thank you! these will be covered in https://github.com/celestiaorg/docs/pull/349 and https://github.com/celestiaorg/devrel-tools/pull/5

PR #406, opened 2023-01-12
Add anchor to link so that link to right position.

### LINKED ISSUE #405: docs: RPC endpoint list link go to wrong position.
On https://docs.celestia.org/nodes/light-node, the RPC endpoint list link of Arabica doesn't use anchor then it goes to the beginning of the article but not the list.

HoytRen:
closes #405

### DIFF
--- a/docs/nodes/light-node.mdx
+++ b/docs/nodes/light-node.mdx
@@ -738,7 +738,7 @@ For ports:
 <Tabs groupId="network">
 <TabItem value="arabica" label="Arabica">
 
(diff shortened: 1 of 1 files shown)


jcstein:
hi @HoytRen and thank you for this PR! Unfortunately, we ran into issues when translating documentation that prevents us from using anchor links. until we've figured out a workaround, I'm going to close this PR and the issue

HoytRen:
Thanks for the info. I will try to help with other issues.

PR #521, opened 2023-03-15
Add a page on metrics for the consensus node

cmwaters:
This PR adds a short walkthough to setting up Prometheus and Grafana to monitor your celestia instance


smuu:
It would be interesting to have a list of available metrics. At the moment, you would need to start everything and look at the Prometheus exporter. When I joined and played with metrics, I sought documentation about that.

YazzyYaz (approved):

PR #526, opened 2023-03-15
Fixing a typo: cohort-one-content.md

pd2112:
During Translation on Crowdin, I found a typo which is I'm going to submit to avoid any confusion for translators.

Note:
Before Correction: 238 ### **Jan 9 - Week 16 - submissions bue**
After Correction: 238 ### **Jan 9 - Week 16 - submissions due**


jcstein:
thank you @pd2112 - great catch 😄

PR #562, opened 2023-03-27
Prompt Scavenger

YazzyYaz:


docs/developers/prompt-scavenger.md
```diff
@@ -0,0 +1,424 @@
...
+
+func main() {
+  // TODO
+}
+```
+
+Here we setup all required libraries we need to use plus the `main`
+function where we will use for our game.
+
+### Helpful Functions
```
jcstein, line 220:
```suggestion
### Helpful functions
```

jcstein (changes requested):
minor edits, mostly on case

PR #570, opened 2023-03-29
Prompt ver bump

YazzyYaz:


jcstein (commented):
I'm going to edit this and add tabs for auto tagged versions, if cool with you

jcstein (approved):
:shipit: i may add tabs in another PR and modify tutorial to allow for different versions

jcstein:
Should we put this in the RPC API section?

PR #623, opened 2023-04-04
docs: add section on docker persistent storage

### LINKED ISSUE #508: Docker section does not handle persistent volumes
https://docs.celestia.org/nodes/docker-images/

This section does not explain handling volumes to have persistent data when restarting the container.

Also: Currently, our container runs with the root user, which will be changed in the future. So, the volume you mount needs to have the correct permission. Ref.: https://github.com/celestiaorg/celestia-app/pull/1473#discussion_r1133758393

I am happy to help with technical details.

jcstein:
## Overview

Resolves #508 

image

## Checklist

### DIFF
--- a/docs/nodes/docker-images.mdx
+++ b/docs/nodes/docker-images.mdx
@@ -270,3 +270,32 @@ you can refer to the
 </TabItem>
 </Tabs>
 ````
+
+## Persistent storage
+
+If you delete a container that you started above, all data will be lost.
+To avoid this, you can mount a volume to the container.
+This will allow you to persist data even after the container is deleted.
+
+Before mounting a volume, you will need to set permissions for
+the user on the host machine by running:
+
+```bash
+sudo chown 10001:10001 ./host/path
+```
+
+In order to mount a volume to the container, you need to specify
+the path to the volume. When you run your container, you can specify
+the path to the volume using the `--volume` (or `-v` for short) flag:
+
+```bash
+# --volume == -v [local path]:[container path]
+docker run [...args] -v ./host/path:/home/celestia celestia <node-type> start [...args]
+```
+
+A full start command will look similar to below. This is an example using
+a light node on Blockspace Race:
+
+```bash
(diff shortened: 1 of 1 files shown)


docs/nodes/docker-images.mdx
```diff
@@ -270,3 +270,32 @@ you can refer to the
...
+
+If you delete a container that you started above, all data will be lost.
+To avoid this, you can mount a volume to the container.
+This will allow you to persist data even after the container is deleted.
+
+Before mounting a volume, you will need to set permissions for
+the user on the host machine by running:
+
+```bash
+sudo chown 10001:10001 ./host/path
```
smuu, line 284: This fails if the specified path does not exist. The user needs to create the folder first.

docs/nodes/docker-images.mdx
```diff
@@ -270,3 +270,32 @@ you can refer to the
...
+```bash
+# --volume == -v [local path]:[container path]
+docker run [...args] -v ./host/path:/home/celestia celestia <node-type> start [...args]
+```
+
+A full start command will look similar to below. This is an example using
+a light node on Blockspace Race:
+
+```bash
+docker run -e NODE_TYPE=light -e P2P_NETWORK=blockspacerace -v /DEV/node:/home/celestia ghcr.io/celestiaorg/celestia-node:v0.8.1 celestia light start --core.ip IP_ADDRESS --gateway --gateway.addr 127.0.0.1 --gateway.port 26659 --p2p.network blockspacerace
```
smuu, line 300: Maybe not use `/DEV/node` for the example.
I propose changing the paths in the whole page to `/tmp/celestia-node-data` so people can follow the tutorial just by copying the commands after each other.
But remember: The `/tmp` folder gets deleted if the host machine is restarted.

smuu (changes requested):

YazzyYaz (approved):

docs/nodes/docker-images.mdx
jcstein: could we use something other than `/tmp` then? I imagine users may not want this to be removed on restart

jcstein: I'm wondering if it is best to provide something like `<your-directory>/celestia-node-data` or something of the sort, so that users need to pay closer attention to the commands. what do you think?

smuu (approved):
nice :rocket:

PR #689, opened 2023-04-26
Simple typo correction

jbroudou:
The documentation tells the reader to create a deployment script named `Contracts.s.sol` however later deployment commands refer to `Contract.s.sol` resulting in a failed deployment.


YazzyYaz (approved):

jcstein (approved):
thank you @jbroudou! good catch 😄

PR #815, opened 2023-06-15
fix: delete redundant subheadings

YazzyYaz:


jcstein (changes requested):
lgtm, I think it'll need to happen in other pages with same instructions (if that's not all)

jcstein:
nice, will plan to replicate this in other instances in the docs too

jcstein (approved):

PR #841, opened 2023-06-28
docs: add guide on celestia-app vesting accounts

### LINKED ISSUE #838: add how to set up vesting account to celestia-app helpful CLI command page

jcstein:
## Overview

Resolves #838 

## Checklist

### DIFF
--- a/docs/developers/celestia-app-vesting.md
+++ b/docs/developers/celestia-app-vesting.md
@@ -1,0 +1,402 @@
+---
+sidebar_label: How to create a vesting account with Celestia App
+description: Learn how to generate a vesting account using Celestia App.
+---
+
+# How to create a vesting account with Celestia App
+
+In this guide, we will learn how to create a vesting account using
+celestia-appd for both a local devnet and on Mocha testnet.
+
+## Local devnet
+
+First, download and [install Celestia App](../../nodes/celestia-app),
+selecting the
+[network and corresponding version](../../nodes/participate)
+that you would like to use.
+
+### Setting up the local devnet
+
+#### Run the devnet
+
+Next, change into the `$HOME/celestia-app` directory and
+run the single-node-devnet script.
+
+```bash
+cd $HOME/celestia-app
+./scripts/build-run-single-node.sh
+```
+
+#### Save the home directory path
+
+At the top of the output, you will see a path to the "Home directory",
+find yours from the output (**it will be unique every time**):
+
+```bash
+./scripts/build-run-single-node.sh
+Home directory: /var/folders/_8/ljj6hspn0kn09qf9fy8kdyh40000gn/T/celestia_app_XXXXXXXXXXXXX.XV92a3qx
+--> Updating go.mod
+```
+
+And set the location as the `NODE_STORE` variable. We will use this
+for the remainder of the devnet section.
+
+```bash
+export NODE_STORE=/var/folders/_8/ljj6hspn0kn09qf9fy8kdyh40000gn/T/celestia_app_XXXXXXXXXXXXX.XV92a3qx
+```
+
+:::tip note
+This does not replace the `celestia-appd` binary that was
+installed with `celestia-appd`, but builds and runs one in the `/build`
+directory.
+:::
+
+#### Check the version of the devnet
+
+If you'd like to check the version of your local devnet, you can use:
+
+```bash
+cd $HOME/celestia-app/build
+./celestia-appd version
+```
+
+#### Next steps
+
+Congratulations! You now have a private devnet running locally on your machine.
+The devnet is made up of one validator that is creating new blocks. This is the
+Celestia consensus network on your machine! The key that was created to run the validator also lives in a temporary directory for the devnet.
+
+Now you are ready to test creating a vesting accout on our devnet before going
+to Mocha, a live testnet.
+
+### Setting up vesting account on devnet
+
+You already have one key setup, but you will need one more to create
+a vesting account.
+
+#### Create a new key
+
+First, create a vesting key:
+
+```bash
+cd $HOME/celestia-app/build
+./celestia-appd keys add vesting-key --home $NODE_STORE
+```
+
+You will see the address, mnemonic, and more details about your key in the
+output:
+
+```bash
+- address: celestia127fpaygehlsgjdknwvlr2mux7h5uvhkxktgkc5
+  name: vesting-key
+  pubkey: '{"@type":"/cosmos.crypto.secp256k1.PubKey","key":"A5JF/we+s5gFt6g944XbKVVYgQB9OY+U/l5dhZjLDczO"}'
+  type: local
+
+
+**Important** write this mnemonic phrase in a safe place.
+It is the only way to recover your account if you ever forget your password.
+
+index enter egg broken ostrich duty bitter blind all car hollow coral youth early verify point void anger daring sausage decline net shove oil
+```
+
+#### List your keys
+
+```bash
+./celestia-appd keys list --home $NODE_STORE
+```
+
+Output:
+
+```bash
+- address: celestia1adgkqcmzuxvg7x5avx8a8rjwpmxgzex3ztef6j
+  name: validator
+  pubkey: '{"@type":"/cosmos.crypto.secp256k1.PubKey","key":"Ahzu6yr9XMPIxLquhgBhj9xL3wIaOz6PE3CvML/oPQym"}'
+  type: local
+- address: celestia127fpaygehlsgjdknwvlr2mux7h5uvhkxktgkc5
+  name: vesting-key
+  pubkey: '{"@type":"/cosmos.crypto.secp256k1.PubKey","key":"A5JF/we+s5gFt6g944XbKVVYgQB9OY+U/l5dhZjLDczO"}'
+  type: local
+```
+
+#### Set variables
+
+Set the keys as varibles, using the validator address as the `FROM_ADDRESS`
+and the vesting-key as the `TO_ADDRESS`.
+
+```bash
+export FROM_ADDRESS=celestia1adgkqcmzuxvg7x5avx8a8rjwpmxgzex3ztef6j
+export TO_ADDRESS=celestia127fpaygehlsgjdknwvlr2mux7h5uvhkxktgkc5
+```
+
+#### Create your devnet vesting account
+
+Create the vesting account with the following command:
+
+:::tip note
+The remainder of the instructions are for a continuous vesting account,
+if you'd like to make a delayed vesting account, use the `--delayed`
+flag.
+:::
+
+```bash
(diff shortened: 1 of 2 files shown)


docs/developers/celestia-app-vesting.md
```diff
@@ -0,0 +1,404 @@
+---
+sidebar_label: How to create a vesting account with Celestia App
```
rootulp, line 2: [not blocking] why can't I find this in the sidebar on PR preview?

I'm looking at https://docs.celestia.org/pr-preview/pr-841/developers/overview/

Screenshot 2023-06-28 at 9 44 46 AM

docs/developers/celestia-app-vesting.md
```diff
@@ -0,0 +1,404 @@
...
+### Setting up the local devnet
+
+#### Run the devnet
+
+Next, change into the `$HOME/celestia-app` directory and
+run the single-node-devnet script.
+
+```bash
+cd $HOME/celestia-app
+./scripts/build-run-single-node.sh
```
rootulp, line 28: [question][discussion] I don't think @celestiaorg/celestia-core maintainers treat scripts located in `./scripts/` as public API. I don't expect any breaking changes to this script in the short term but just flagging that we should probably be careful to not introduce breaking changes to any of the scripts that are referenced from the docs.

docs/developers/celestia-app-vesting.md
```diff
@@ -0,0 +1,404 @@
...
+Home directory: /var/folders/_8/ljj6hspn0kn09qf9fy8kdyh40000gn/T/celestia_app_XXXXXXXXXXXXX.XV92a3qx
+--> Updating go.mod
+```
+
+And set the location as the `NODE_STORE` variable. We will use this
+for the remainder of the devnet section.
+
+```bash
+export NODE_STORE=/var/folders/_8/ljj6hspn0kn09qf9fy8kdyh40000gn/T/celestia_app_XXXXXXXXXXXXX.XV92a3qx
```
rootulp, line 46: [nit] is it possible to use a different variable name? `NODE_STORE` makes me think this is related to celestia-node but this variable is for the celestia-app home.

When I do this locally, I usually name it `CELESTIA_APP_HOME` but up to you.

docs/developers/celestia-app-vesting.md
```diff
@@ -0,0 +1,404 @@
...
+```
+
+:::tip note
+This does not replace the `celestia-appd` binary that was
+installed with `celestia-appd`, but builds and runs one in the `/build`
```
rootulp, line 51: [nit] to clarify that the build directory referenced is local to the celestia-app repo and not some global `/build` directory

```suggestion
installed with `celestia-appd`, but builds and runs one in the `./build`
```

docs/developers/celestia-app-vesting.md
rootulp: ^ disregard found it in https://docs.celestia.org/pr-preview/pr-841/developers/celestia-app-vesting/

after seeing the last file in this Github PR

rootulp (approved):

docs/developers/celestia-app-vesting.md
jcstein: yea, it's hidden by default. do you think it would it be better if it is open by default?

docs/developers/celestia-app-vesting.md
jcstein: definitely possible, i'll use `CELESTIA_APP_HOME` thanks!

docs/developers/celestia-app-vesting.md
jcstein: good point, i'm not sure here actually. do you think it is a good idea to just put a script in to the docs for people to copy on their own?

docs/developers/celestia-app-vesting.md
jcstein: resolved in 8adf1efdb71f53871914f5fa87054a0dc4c6e216

docs/developers/celestia-app-vesting.md
jcstein: added full path in this line for clarity in d1e148b5563e4b7e04a11f9e2d6c5917bab7ace6

docs/developers/celestia-app-vesting.md
rootulp: I don't have a strong opinion. TBH I wasn't expecting it to be under the "Wallet with Celestia App" category but idk where it belongs so defer to you.

docs/developers/celestia-app-vesting.md
rootulp: hmm if we put the script into the docs then it may become stale / break due to changes in celestia-app. I think I'd rather the script live in the celestia-app repo.

If we really want to ensure the docs don't break, we could introduce a CI test that tries to run this script. Or alternatively, we could write a new script that contains the steps on this docs page that runs in CI so we don't break it accidentally. Thoughts?

jcstein: if it's worth the testing resources, I think it would be worth it. at the moment though, not sure where it lies in terms of scope/bandwidth. so will leave this PR open for now and continue to think on it

docs/developers/celestia-app-vesting.md
jcstein: any ideas? i didn't want to bury it in resources, but it may fit better there

docs/developers/celestia-app-vesting.md
```diff
@@ -0,0 +1,404 @@
...
+
+```bash
+cd $HOME/celestia-app
+./scripts/build-run-single-node.sh
+```
+
+#### Save the home directory path
+
+At the top of the output, you will see a path to the "Home directory",
+find yours from the output (**it will be unique every time**):
```
staheri14, line 34: [suggestion, optional] It would be helpful if we could add a sample output (we could conceal the output path though, so that users won't mistakenly use it)

docs/developers/celestia-app-vesting.md
jcstein: expanding menu for now by default, i can't think of another way atm or better place

docs/developers/celestia-app-vesting.md
```diff
@@ -0,0 +1,404 @@
+---
+sidebar_label: How to create a vesting account with Celestia App
+description: Learn how to generate a vesting account using Celestia App.
+---
+
+# How to create a vesting account with Celestia App
+<!-- markdownlint-disable MD013 -->
+
+In this guide, we will learn how to create a vesting account using
+celestia-appd for both a local devnet and on Mocha testnet.
```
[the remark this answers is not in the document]
jcstein: thank you! resolved in bd09b29f19ce5d54a3337c74a8fea81c71c8b256

docs/developers/celestia-app-vesting.md
jcstein: sample output like this? https://github.com/celestiaorg/docs/blob/bd09b29f19ce5d54a3337c74a8fea81c71c8b256/docs/developers/celestia-app-vesting.md?plain=1#L31-L47

how would you recommend concealing the path? i think having the full example makes it easier for someone _who is actually reading the guide_ to understand what type of path format to look for?

docs/developers/celestia-app-vesting.md
jcstein: i noticed that the single node script is in the readme. is there ci for that there? maybe it is possible to add a comment in the readme that says "TODO: if you delete this, update docs too" by the node script?

docs/developers/celestia-app-vesting.md
staheri14: > sample output like this?

Right! 👍 

> how would you recommend concealing the path? 

I think the current sample looks good.

jcstein: okay, thank you for the feedback! open to suggestions 🙏

YazzyYaz (commented):
Can I have more context on why we are switching nomenclature from:
Consensus Full Nodes to Full Consensus Nodes?

Reason it was architected this way is because the Full Storage Node exists as a nomenclature so this was to make both less confusing.

If we want to do Full Consensus Nodes, imo we should then rename Full Storage Node to Full DA Node or just call it simply a Storage Node. Otherwise it'll get too confusing between different node types.

To not block this PR, I would recommend we open a new issue to discuss name changes so it doesn't block this PR from being merged.

docs/developers/celestia-app-vesting.md
rootulp: Unfortunately there isn't CI that ensures our README commands run as expected. I can add a TODO in celestia-app repo if we want to merge this as-is

rootulp: Created https://github.com/celestiaorg/celestia-app/issues/2007 as a potential longer term solution

jcstein:
hey @YazzyYaz, i switched up the nomenclature in the docs in #844, because as I was reading through the celestia-app repo I noticed it was called that way so made #843 for it.

I have had a hard time understanding what type of full node people are referring to since I've been working with all things Celestia. So my thinking was that instead of "consensus full node" and "full storage node", where "consensus" usually gets dropped from "consensus full node", leaving "full node" to be confusing -- we title them "full consensus node" and "full storage node" so that full and node are always tied with the third piece that further clarifies what type of full node it is. That, and matching celestia-app's repo.

Will resolve conflict, but can also revert #844 if you think it is best

jcstein:
that being said, depending on what you think is best to keep, i think if using "full consensus node" we should instead call "full storage node" "full DA node" like you proposed

docs/developers/celestia-app-vesting.md
adlerjohn: If the scripts are needed to run as an end-user, then they're part of the app's public-facing API and should be 1) documented and 2) tested in CI.

SweeXordious (commented):
:tada: :+1: 

[Non-blocking][Question]
Does it make sense to specify which kind of vesting account this tutorial is creating?

Also, not all vesting accounts can be created with a message, [source](https://docs.cosmos.network/v0.46/modules/auth/05_vesting.html#note), some need to be set at genesis. Does  it make sense to mention this?

jcstein:
yes, i think it makes sense to clarify @SweeXordious - can you help me understand what type I created 😅 ? thank you!

SweeXordious:
@jcstein Sure, it's creating a *continuous vesting account*.
You could also create a *delayed vesting account* via adding the `--delayed` flag to the command: 
```
./celestia-appd tx vesting create-vesting-account $TO_ADDRESS 100000utia 1686748051 --from $FROM_ADDRESS --gas auto --fees 100000utia --chain-id private --home $CElESTIA_APP_HOME -y
```

jcstein:
got it! i had this note in the guide https://github.com/celestiaorg/docs/blob/0372c3434d8ec4fbda1a4fef7a1d2f844267e5b2/docs/developers/celestia-app-vesting.md?plain=1#L137-L141, but will work on clarifying further in the guide

PR #862, opened 2023-07-05
doc: Bump to Mocha 3

YazzyYaz:


jcstein (changes requested):
I don't think we need this PR, unless the way celestia-node handles aliases has changed

for example, in the past, when we bump arabica from `arabica-2` to `arabica-3` the alias is changed in celestia-node itself, so the command `--p2p.network arabica` refers to whichever the live network is. the user should not need to worry about the chain ID

so i don’t think we need to add `mocha-3` in places where the alias `mocha` is used?

jcstein:
Additionally, on v0.11.0-rc6 of celestia-node, if the `mocha-3` flag is passed, it returns an error:

```bash
celestia light init --p2p.network mocha-3
Error: invalid network specified: mocha-3
Usage:
  celestia light init [flags]
```

jcstein:
closing in favor of https://github.com/celestiaorg/docs/pull/868

PR #957, opened 2023-08-17
docs: update instantiate-testnet.md guide

### LINKED ISSUE #956: issue with local testnet guide
from a user:
For me the most confusing part is "Add genesis account KeyName", where command uses $KEY_NAME but sentence after "Here $VALIDATOR_NAME is the same key name as before; and $AMOUNT is something like 10000000000000000000000000utia. " throws me off

1. it should be fixed to say: 
Here $KEY_NAME...

2. experiencing a bug when creating gentx: 
```bash
STAKING_AMOUNT=1000000000utia
celestia-appd gentx $KEY_NAME $STAKING_AMOUNT --chain-id $CHAIN_ID \
  --keyring-backend test
Error: failed to build create-validator message: the provided eth address is not a valid hex address
```

jcstein:
## Overview

Resolves #956 

## Checklist

### DIFF
--- a/docs/nodes/instantiate-testnet.md
+++ b/docs/nodes/instantiate-testnet.md
@@ -76,7 +76,7 @@ CELES_AMOUNT="10000000000000000000000000utia"
 celestia-appd add-genesis-account $KEY_NAME $CELES_AMOUNT --keyring-backend test
 ```
 
-Here `$VALIDATOR_NAME` is the same key name as before; and `$AMOUNT`
+Here `$KEY_NAME` is the same key name as before; and `$AMOUNT`
 is something like `10000000000000000000000000utia`.
 
 ### Optional: Adding other validators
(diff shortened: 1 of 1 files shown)


docs/nodes/instantiate-testnet.md
```diff
@@ -76,7 +76,7 @@ CELES_AMOUNT="10000000000000000000000000utia"
 celestia-appd add-genesis-account $KEY_NAME $CELES_AMOUNT --keyring-backend test
 ```
 
-Here `$VALIDATOR_NAME` is the same key name as before; and `$AMOUNT`
+Here `$KEY_NAME` is the same key name as before; and `$AMOUNT`
 is something like `10000000000000000000000000utia`.
```
rootulp, line 80: [nit][not blocking] the $AMOUNT flag appears to reference $CELES_AMOUNT and that's already re-defined in the code above so it doesn't appear to need extra explanation.

```suggestion
Here `$KEY_NAME` is the same key name as before.
```

rootulp: This looks correct based on https://github.com/celestiaorg/celestia-app/blob/e2b154d8314872eba6f0d8d5ea18fdb58b201150/scripts/single-node.sh#L16-L17

where $KEY_NAME is defined as validator above. It may make sense to set all the environment variables at the top of this file but just a thought

rootulp (approved):

rootulp (approved):

Issue #556, opened 2023-02-01
Consider consolidating consensus and validator pages on docs


evan-forbes:

jcstein:
Hm, do we want to leave these node types separate as one is for app and one is for node?

rootulp:
https://docs.celestia.org/nodes/validator-node/ and https://docs.celestia.org/nodes/consensus-full-node/ both use celestia-app.

jcstein:
I think my comment above is a good example of the confusion introduced by naming different node types:
- celestia-node light node
- celestia-node full node
- celestia-node bridge node
- celestia-app consensus validator node
- celestia-app consensus full node

The celestia-app consensus full node and celestia-node full node are often both referred to as "full nodes" In an effort to clear confusion, I opened #843 and #844 to rename the celestia-app consensus full node in documentation to match the GitHub repository of celestia-app. The "consensus full node" is now referred to as "full consensus node"

I will still consider consolidating the validator and consensus full node pages, but am not sure exactly how so will revisit in the coming weeks

jcstein:
<img width="299" alt="scrot 2023-09-29 at 10 09 08 PM" src="https://github.com/celestiaorg/docs/assets/46639943/1cd36bad-b70c-4586-8d4c-9003d0be3287">

PR #1095, opened 2023-09-28
chore: resolve inaccuracies around exposing consensus node's RPC

### LINKED ISSUE #522: Inaccuracies around exposing consensus node's RPC
In the docs under the consensus node. There is a section https://docs.celestia.org/nodes/consensus-full-node/#optional-configure-for-rpc-endpoint which I don't believe is quite accurate

> ### Optional: configure for RPC endpoint
>You can configure your Consensus Full Node to be a public RPC endpoint and listen to any connections from Data Availability > >Nodes in order to serve requests for the Data Availability API [here](https://docs.celestia.org/developers/node-tutorial/).
> 
>Note that you would need to ensure port 9090 is open for this.
>
>Run the following commands:
>
>EXTERNAL_ADDRESS=$(wget -qO- eth0.me)
>sed -i.bak -e "s/^external-address = \"\"/external-address = \"$EXTERNAL_ADDRESS:26656\"/" $HOME/.celestia-app/config/config.toml
>sed -i 's#"tcp://127.0.0.1:26657"#"tcp://127.0.0.1:26657"#g' ~/.celestia-app/config/config.toml
>
>Restart celestia-appd in the previous step to load those configs.

Populating the external address isn't necessary to making the RPC endpoint public. The `external_address` (which uses an underscore instead of a hyphen) is commonly for when you're running your node behind a NAT and so you need to advertise a different address for peers to dial. 

> sed -i 's#"tcp://127.0.0.1:26657"#"tcp://127.0.0.1:26657"#g' ~/.celestia-app/config/config.toml

This part doesn't seem to change anything. You could suggest using the public IP instead of localhost to expose the port directly or we could include documentation around setting up a reverse proxy (which I think is the standard way people expose RPC).

> Note that you would need to ensure port 9090 is open for this.

I was a bit confused about this at first because 9090 is not normally the port for RPC. As I understand it, this refers to making gRPC public as it's also required by bridge nodes. Could we better clarify this.

jcstein:
note: after this is implemented, i plan to complete #556

## Overview

Resolves #522 

## Checklist

### DIFF
--- a/docs/nodes/full-consensus-node.mdx
+++ b/docs/nodes/full-consensus-node.mdx
@@ -242,7 +242,7 @@ which ports are required to be open on your machine.
 ### Optional: Configuring an RPC endpoint
 
 You can configure your Consensus Full Node to be a public RPC endpoint.
-This allows it to accept connections from Data Availability Nodes and]
+This allows it to accept connections from Data Availability Nodes and
 serve requests for the Data Availability API.
 
 #### Expose RPC
--- a/docs/nodes/full-consensus-node.mdx
+++ b/docs/nodes/full-consensus-node.mdx
@@ -239,21 +239,44 @@ Please refer to the [ports](../../nodes/celestia-node/#ports) section for inform
 which ports are required to be open on your machine.
 :::
 
-### Optional: configure for RPC endpoint
+### Optional: Configuration for RPC endpoint
 
-You can configure your full consensus node to be a public RPC endpoint
-and listen to any connections from Data Availability Nodes in order to
-serve requests for the Data Availability API [here](../developers/node-tutorial.mdx).
+You can configure your Consensus Full Node to be a public RPC endpoint.
+This allows it to accept connections from Data Availability Nodes and]
+serve requests for the Data Availability API.
 
-Run the following commands:
+#### Expose RPC
+
+By default, the RPC service listens on `localhost` which means it can't
+be accessed from other machines. To make the RPC service available publicly,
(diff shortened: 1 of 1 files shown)


docs/nodes/full-consensus-node.mdx
```diff
@@ -239,21 +239,44 @@ Please refer to the [ports](../../nodes/celestia-node/#ports) section for inform
 which ports are required to be open on your machine.
 :::
 
-### Optional: configure for RPC endpoint
+### Optional: Configuration for RPC endpoint
```
jcstein, line 242:
```suggestion
### Optional: Configuring an RPC endpoint
```

docs/nodes/full-consensus-node.mdx
```diff
@@ -239,21 +239,44 @@ Please refer to the [ports](../../nodes/celestia-node/#ports) section for inform
...
 :::
 
-### Optional: configure for RPC endpoint
+### Optional: Configuring an RPC endpoint
 
-You can configure your full consensus node to be a public RPC endpoint
-and listen to any connections from Data Availability Nodes in order to
-serve requests for the Data Availability API [here](../developers/node-tutorial.mdx).
+You can configure your Consensus Full Node to be a public RPC endpoint.
+This allows it to accept connections from Data Availability Nodes and]
```
jcstein, line 245:
```suggestion
This allows it to accept connections from Data Availability Nodes and
```

cmwaters (approved):
LGTM. Thanks for doing this

jcstein:
you're welcome, thank you for the descriptive issue!

PR #1184, opened 2023-10-24
docs: improve hermes `CompatMode` tip

rootulp:
Now that hermes v1.7.0 has been released, we can recommend an official hermes release instead of a patch.


jcstein (approved):
LGTM, thank you for the PR and context!

PR #1219, opened 2023-10-30
fix: validator updates

jcstein:
## Overview

[Preview of consensus node page](https://docs.celestia.org/pr-preview/pr-1219/nodes/consensus-node#storage-and-pruning-configurations)

## Checklist

### DIFF
--- a/nodes/consensus-node.md
+++ b/nodes/consensus-node.md
@@ -169,7 +169,7 @@ indexer = "null"
 
 ```toml
 # by setting custom and a keep recent value of 2, the node will only 
-# keep the past two state snapshotss
+# keep the past two state snapshots
 pruning = "custom"
 pruning-keep-recent = "2"
 pruning-interval = "10"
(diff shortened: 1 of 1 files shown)


nodes/consensus-node.md
```diff
@@ -151,180 +151,63 @@ alias head=ghead
...
-
-```toml
-indexer = "kv" # or "psql"
-```
-
-`app.toml`:
-
-```toml
-min-retain-blocks = 0
+indexer = "kv"
```
evan-forbes, line 163: let's keep this for now, but I think there was a miscommunication somewhere. I was under the impression that this was needed for bridge nodes, but after looking through the code I don't think it is. There's no time to properly test if it is or not, so I think its best to keep this as is for now.

evan-forbes (approved):
:shipit:

nodes/consensus-node.md
jcstein: understood, thank you for the context.

PR #1237, opened 2023-11-02
fix: fix docker start command

### LINKED ISSUE #1228: Docker Light Node fails to start second time and Docker Docs Are Incorrect for Start
The example command for starting the node is the same as what is used to init as seen here:

https://docs.celestia.org/nodes/docker-images#start-the-node

Additionally, you can just start the node to begin with and it will run the init anyways, perhaps the current process is just to simplify keeping track of the address and mnemonic.

The first time the docker container is started and running it appears to function properly, but once stopped if you try to restart the container or create a new container referencing the same volume the client complains about permissions on the file.

```
2023-10-31 13:03:32 Error: could not build arguments for function "github.com/celestiaorg/celestia-node/nodebuilder/p2p".ConstructModule.Listen.func2 (/src/nodebuilder/p2p/addrs.go:14): failed to build host.Host: could not build arguments for function "github.com/celestiaorg/celestia-node/nodebuilder/p2p".routedHost (/src/nodebuilder/p2p/host.go:38): failed to build p2p.HostBase: could not build arguments for function "github.com/celestiaorg/celestia-node/nodebuilder/p2p".host (/src/nodebuilder/p2p/host.go:43): failed to build peer.ID: could not build arguments for function "github.com/celestiaorg/celestia-node/nodebuilder/p2p".id (/src/nodebuilder/p2p/identity.go:45): failed to build crypto.PrivKey: received non-nil error from function "github.com/celestiaorg/celestia-node/nodebuilder/p2p".Key (/src/nodebuilder/p2p/identity.go:17): keystore: permissions of key 'p2p-key' are too relaxed: required: 0600, got: 0777
```

jcstein:
## Overview

Resolves #1228
[running with persistent storage logs](https://app.warp.dev/block/qXtesjeJ8kLiWfS3FVDahN)

## Checklist

### DIFF
--- a/nodes/docker-images.md
+++ b/nodes/docker-images.md
@@ -76,8 +76,9 @@ Ubuntu. You can
 
    :::
 
-3. Set an RPC endpoint for either [Arabica](./arabica-devnet.md#rpc-endpoints) or
-   [Mocha](./mocha-testnet.md#rpc-endpoints)
+3. Set an RPC endpoint for either [Mainnet Beta](./mainnet.md#da-rpc-endpoints),
+   [Mocha](./mocha-testnet.md#rpc-endpoints), or
+   [Arabica](./arabica-devnet.md#rpc-endpoints)
    using the bare URL (without http or https):
 
    ```bash
--- a/nodes/docker-images.md
+++ b/nodes/docker-images.md
@@ -36,7 +36,10 @@ Ubuntu. You can
 
 ## Prerequisites
 
-- [Docker Desktop](https://docs.docker.com/get-docker) and basic understanding of Docker
+- [Docker Desktop for Mac or Windows](https://docs.docker.com/get-docker) and a basic
+understanding of Docker
+- [Docker Engine for Linux](https://docs.docker.com/engine/install/) and a
+basic understanding of Docker
 
 ## Quick start
 
@@ -143,7 +146,7 @@ the user on the host machine by running:
 sudo chown 10001:10001 $HOME/my-node-store
 ```
 
-```bash [Mac]
+```bash [Docker Desktop on Mac]
 # you're good to go 😎
 ```
 
--- a/nodes/docker-images.md
+++ b/nodes/docker-images.md
@@ -151,7 +151,7 @@ using the variables we set in the [quick start](#quick-start) section:
 
 ```bash
 # --volume == -v [local path]:[container path]
-docker run [args...] -v $HOME/my-node-store:/celestia \
+docker run [args...] -v $HOME/my-node-store:/home/celestia \
     celestia $NODE_TYPE init [args...]
 ```
 
@@ -161,21 +161,21 @@ An example init command will look similar to below:
 
 ```bash-vue [Mainnet Beta]
 docker run -e NODE_TYPE=$NODE_TYPE -e P2P_NETWORK=$NETWORK \
-    -v $HOME/my-node-store:/celestia \
+    -v $HOME/my-node-store:/home/celestia \
     ghcr.io/celestiaorg/celestia-node:{{mainnetVersions['node-latest-tag']}} \
     celestia light init --p2p.network $NETWORK
 ```
 
 ```bash-vue [Mocha]
 docker run -e NODE_TYPE=$NODE_TYPE -e P2P_NETWORK=$NETWORK \
-    -v $HOME/my-node-store:/celestia \
+    -v $HOME/my-node-store:/home/celestia \
     ghcr.io/celestiaorg/celestia-node:{{mochaVersions['node-latest-tag']}} \
     celestia light init --p2p.network $NETWORK
 ```
 
 ```bash-vue [Arabica]
 docker run -e NODE_TYPE=$NODE_TYPE -e P2P_NETWORK=$NETWORK \
-    -v $HOME/my-node-store:/celestia \
+    -v $HOME/my-node-store:/home/celestia \
     ghcr.io/celestiaorg/celestia-node:{{arabicaVersions['node-latest-tag']}} \
     celestia light init --p2p.network $NETWORK
 ```
@@ -188,7 +188,7 @@ Run the following command to start the node:
 
 ```bash
 # --volume == -v [local path]:[container path]
-docker run [...args] -v $HOME/my-node-store:/celestia \
+docker run [...args] -v $HOME/my-node-store:/home/celestia \
     celestia <node-type> start [...args]
 ```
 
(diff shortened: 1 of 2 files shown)


nodes/docker-images.md
```diff
@@ -76,8 +76,9 @@ Ubuntu. You can
...
    :::
 
-3. Set an RPC endpoint for either [Arabica](./arabica-devnet.md#rpc-endpoints) or
-   [Mocha](./mocha-testnet.md#rpc-endpoints)
+3. Set an RPC endpoint for either [Mainnet Beta](./mainnet.md#da-rpc-endpoints),
+   [Mocha](./mocha-testnet.md#rpc-endpoints), or
+   [Arabica](./arabica-devnet.md#rpc-endpoints)
    using the bare URL (without http or https):
 
    ```bash
```
keyneom, line 84: It feels like this next line would ideally provide a sample command with a real RPC_URL (or a few different ones) across the different networks as is done for the other commands (e.g. 4 from below let's you choose the network you are trying to run on and provides the sample command to use in that case).

nodes/docker-images.md
```diff
@@ -160,21 +161,21 @@ An example init command will look similar to below:
 
 ```bash-vue [Mainnet Beta]
 docker run -e NODE_TYPE=$NODE_TYPE -e P2P_NETWORK=$NETWORK \
```
keyneom, line 163: This should be added here:
```
-e CELESTIA_HOME=/celestia
```

or a section above should be added exporting a variable that can be used here like the others.

Or the other option would be to modify the default docker environment variable to use the `/celestia` path instead of `/home/celestia`.

Or you may want to switch the other commands back to using `/home/celestia` for the mounted volume instead of `/celestia` to maintain persistence.

nodes/docker-images.md
keyneom, line 1: I'd make a note that trying to run the docker images in windows is likely to run into issues due to the way NTFS handles permissions and that referencing a volume that supports linux permissions (such as within WSL2) is a good way for Windows users to be able to run a node with persistent storage.

keyneom (changes requested):

nodes/docker-images.md
jcstein: please feel free to make a PR to add the instructions that work for you on Windows!

nodes/docker-images.md
jcstein: i will not provide a sample command with a sample URL as to allow the user to pick their own

nodes/docker-images.md
jcstein: i tested the full command when i edited this doc and it worked as expected. `-v $HOME/my-node-store:/celestia \` does the trick

jcstein: what OS are you on? i think that may be the issue. this doc works for mac and linux.

keyneom: I've executed it both within windows and Ubuntu WSL. Are you sure it is actually persisting in your testing? The command executes successfully as written for me as well, it just doesn't actually persist anything. So after running it you can check in the `$HOME/my-node-store` folder on your host machine and find no files or directories were created (of course you should wipe the directory first to ensure it is empty to begin with and isn't using the data from previous runs).

So while I think all of the commands in the docs will not currently throw any errors, I do believe they are not correct as currently written as they will not achieve the intended goal of a persistent node. Perhaps I messed up somehow though.

nodes/docker-images.md
keyneom: For sure that's totally reasonable. I do think--like I mentioned above--that the current commands do not achieve the intended objective of persisting a node although they may not throw any errors but perhaps you are referencing a different version of the docker container or something like that which works.

keyneom: @jcstein I created this PR https://github.com/celestiaorg/docs/pull/1249 feel free to modify/improve wording as much as you feel is appropriate.

jcstein: thank you! taking a look at the pr and comment above

nodes/docker-images.md
jcstein: got it, i’ll test to verify!

nodes/docker-images.md
jcstein: it looks like it isn't persisting. @smuu could you please help take a look at this?

jcstein: [Preview of section](https://docs.celestia.org/pr-preview/pr-1237/nodes/docker-images#light-node-setup-with-persistent-storage)

keyneom: The additional parameter I recommend adding to the docker commands should fix this so that it does persist correctly. That, or the other options I mentioned in the other comment thread above should work as well.

nodes/docker-images.md
keyneom: The original comment in this thread was to fix the persistence issue I had encountered.

nodes/docker-images.md
jcstein: Understood! What is the full start command then? As I understand the `-v $HOME/my-node-store:/home/celestia` flag would still be used?

jcstein:
```bash
docker run -e NODE_TYPE=$NODE_TYPE -e P2P_NETWORK=$NETWORK -e CELESTIA_HOME=/celestia -v $HOME/my-node-store:/celestia ghcr.io/celestiaorg/celestia-node:v0.12.0 celestia $NODE_TYPE init --p2p.network $NETWORK
Initializing Celestia Node with command:
celestia light init --p2p.network celestia
2023-11-09T10:36:03.394Z	INFO	node	nodebuilder/init.go:31	Initializing Light Node Store over '/celestia/.celestia-light'
Error: mkdir /celestia/.celestia-light: operation not permitted
```

jcstein: it appears to be something related to permissions in the container itself afaiu

jcstein: after changing the path back to `/home/celestia` and using your `-e CELESTIA_HOME=/home/celestia` flag, [it works!](https://app.warp.dev/block/sPjiKSEWP2uwQHI6M5TpFN)

smuu:
So my findings:
- when using `/home/celestia` instead of `/celestia`, the keys and the data are persisted (the node uses this if not specified otherwise
- when using Mainnet Beta, we don't need to set `P2P_NETWORK` as this points by default to the `celestia` network
- `NODE_TYPE` is not needed at all, I think, as this is only used when automatically calling init when starting, but as we have persistent data and we init manually, this is not needed in this section

This should resolve the bug and make this section a little cleaner.

nodes/docker-images.md
jcstein: after changing the path back to `/home/celestia` and using your `-e CELESTIA_HOME=/home/celestia` flag, [i hit an error not being able to create the directory in the container](https://app.warp.dev/block/NkhVSyrTKXE9WBd2LFe7BN)?

jcstein:
[copying this](https://github.com/celestiaorg/docs/pull/1237#discussion_r1387926963) for vis @smuu 

after changing the path back to /home/celestia and using your -e CELESTIA_HOME=/home/celestia flag, [i hit an error not being able to create the directory in the container](https://app.warp.dev/block/NkhVSyrTKXE9WBd2LFe7BN)?

jcstein:
i plan to use p2p network and node type so that users can set up env once in above section in order to not have to explain/set the same variables twice

smuu:
> i plan to use p2p network and node type so that users can set up env once in above section in order to not have to explain/set the same variables twice

I understand, then I propose to use `$NODE_TYPE` also in the init and start commands, as those are hardcoded to `light` at the moment.

smuu:
> [copying this](https://github.com/celestiaorg/docs/pull/1237#discussion_r1387926963) for vis @smuu
> 
> after changing the path back to /home/celestia and using your -e CELESTIA_HOME=/home/celestia flag, [i hit an error not being able to create the directory in the container](https://app.warp.dev/block/NkhVSyrTKXE9WBd2LFe7BN)?

I can't reproduce this behavior.
When following the steps of creating the folder and changing the permissions, I can run the container with the init.

keyneom:
> > [copying this](https://github.com/celestiaorg/docs/pull/1237#discussion_r1387926963) for vis @smuu
> > after changing the path back to /home/celestia and using your -e CELESTIA_HOME=/home/celestia flag, [i hit an error not being able to create the directory in the container](https://app.warp.dev/block/NkhVSyrTKXE9WBd2LFe7BN)?
> 
> I can't reproduce this behavior. When following the steps of creating the folder and changing the permissions, I can run the container with the init.

That is my experience as well. As long as the permissions are set with the prior command (`chown`) then I don't encounter any errors and things persist correctly.

keyneom:
> So my findings:
> 
> * when using `/home/celestia` instead of `/celestia`, the keys and the data are persisted (the node uses this if not specified otherwise
> * when using Mainnet Beta, we don't need to set `P2P_NETWORK` as this points by default to the `celestia` network
> * `NODE_TYPE` is not needed at all, I think, as this is only used when automatically calling init when starting, but as we have persistent data and we init manually, this is not needed in this section
> 
> This should resolve the bug and make this section a little cleaner.

I believe the docs as written are intended to support running on test nets, etc so I'd leave the `P2P_NETWORK` as is. I do think it makes sense to just use `/home/celestia` since it is the default. I wasn't clear on the other factors that may have been driving a switch to `/celestia`.

jcstein:
I think we're in good shape now!

smuu (approved):
lgtm

PR #1249, opened 2023-11-03
Persistent Docker Node Instructions for Windows Users

keyneom:
## Overview

This PR provides some information necessary for Windows users to be able to successfully run a persistent docker node. NTFS does not support linux style permissions so when a docker node is started the first time it creates the desired files to store the node, but afterwards will refuse to run indicating that the file containing the keys has permissions of 777 instead of 600. In order to avoid these issues it is recommended that Windows users install and run WSL and then run the docker container with the celestia node storage mounted within the WSL filesystem which obviously supports linux permissions.

Associated issue can be found here #1228 .

## Checklist

- [X] New and updated code has appropriate documentation
- [X] New and updated code has new and/or updated testing
- [X] Required CI checks are passing
- [X] Visual proof for any user facing features like CLI or documentation updates

### DIFF
--- a/nodes/docker-images.md
+++ b/nodes/docker-images.md
@@ -122,11 +122,13 @@ If you delete a container that you started above, all data will be lost.
 To avoid this, you can mount a volume to the container.
 This will allow you to persist data even after the container is deleted.
 
-For security purposes Celestia expects to interact with the your node's keys in a read-only manner.
-This is enforced using linux style permissions on the filesystem.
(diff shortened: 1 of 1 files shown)


smuu (commented):
Good catch :+1: 

This needs to be more highlighted.

Many users (including myself) skip reading the text and only copy & paste the commands.

jcstein (approved):

PR #1301, opened 2023-11-29
Update dev docs for L2s

YazzyYaz:


musalbas:
Can you also rename to Ethereum fallback rather than ETH fallback?

musalbas:
It's still titled ETH fallback on the page: https://docs.celestia.org/developers/eth-fallback

https://github.com/celestiaorg/docs/issues/1302

Issue #1326, opened 2023-12-14
Add docs for upgrade monitor tool


rootulp:
## Context

We wrote a tool that monitors upgrades on celestia-app and optionally auto-publishes a signed transaction. We ideally want at least one person to run the tool during an upgrade.

## Proposal

After we merge the tool to celestia-app or a stand-alone repo, we should add to docs a short description of the tool + link so that it is discoverable.

rootulp:
We merged https://github.com/celestiaorg/upgrade-monitor so I can add a short blurb about it in the docs.

PR #1329, opened 2023-12-14
docs: remove Blockspace Race section

rootulp:
This section seems no longer necessary considering Blockspace Race is over.


jcstein (approved):
agreed, good catch. more context on upgrade tool can be added for #1326

PR #1323, opened 2023-12-12
feat: add initial arbitrum docs

jcstein:
## Overview

Adds documentation for Arbitrum:

1. [Introduction to Arbitrum rollups with Celestia as DA](https://celestiaorg.github.io/docs/pr-preview/pr-1323/developers/arbitrum-integration)
2. [Deploy an Arbitrum rollup devnet](https://celestiaorg.github.io/docs/pr-preview/pr-1323/developers/arbitrum-deploy)
3. [Deploy a smart contract to your Arbitrum rollup](https://celestiaorg.github.io/docs/pr-preview/pr-1323/developers/arbitrum-smart-contract)
4. [Deploy a dapp on your Arbitrum rollup](https://celestiaorg.github.io/docs/pr-preview/pr-1323/developers/arbitrum-dapp-deploy)

Depends on

- [x] https://github.com/celestiaorg/nitro-testnode/pull/4
- [x] https://github.com/celestiaorg/nitro/pull/2
- [x] remove the workarounds once these are merged

This PR also resolves #1303, restructuring the Ethereum fallback page.

## Checklist

- [x] New and updated code has appropriate documentation
- [x] New and updated code has new and/or updated testing
- [x] Required CI checks are passing
- [x] Visual proof for any user facing features like CLI or documentation updates
- [x] Linked issues closed with keywords

PR #1332, opened 2023-12-14
chore: run yarn lint

rootulp:
While working on a different PR, the commit pre-hook failed because `yarn lint` resulted in a diff. Since these changes are unrelated to my PR, I included them here. 

Note: I wonder why CI lint didn't catch these? Do we want an issue to investigate the discrepancy between local lint and CI lint?

### DIFF
--- a/developers/blobstream.md
+++ b/developers/blobstream.md
@@ -99,11 +99,13 @@ You can interact with the Blobstream contracts today on testnet. The Blobstream 
 smart contracts are currently deployed on the following Ethereum testnets:
 
 <!-- markdownlint-disable MD013 -->
-| Contract | EVM network | Contract address | Attested data |
-| --- | --- | --- | --- |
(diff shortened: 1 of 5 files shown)


jcstein:
I had this happen to me and added the linting in a PR I have open (not best practice), so thanks for opening this. Just [reverted the commit](https://github.com/celestiaorg/docs/pull/1323/commits/f033b2f5d410cae8fbd433cfe91d9a0c578f397a) on that PR.

> Note: I wonder why CI lint didn't catch these? Do we want an issue to investigate the discrepancy between local lint and CI lint?

It doesn't look like there is a CI lint on the most recent PRs...so looking into that now.

Regardless, I think that we do need to look into the flow of the linting with the pre-commit hook, and whether it should also exist the same way as CI, as I don't see it running on this PR? As for now, it only is used if developer downloads dependencies, otherwise husky won't run the pre-commit

jcstein (approved):

PR #1489, opened 2024-03-20
docs: add documentation on how to build rollups on Celestia with blobstream

rach-id:
## Overview

Given the number of people asking how to build a rollup on Celestia that uses Blobstream, this document aims to fill that lack.

## Checklist

## Summary by CodeRabbit

- **Documentation**
	- Updated guidance on building rollups using Blobstream, focusing on optimistic rollups and Zk-rollups, including share commitments and sequence of spans for data availability.
	- Added a new section on "Blobstream rollups" with a reference to additional documentation.
	- Added a new sidebar item titled "Blobstream rollups" with a link to "/developers/blobstream-rollups".


developers/blobstream-rollups.md
```diff
@@ -0,0 +1,54 @@
...
+
+So, when submitting the headers to the rollup settlement contract, the settlement contract will need to verify the following:
+
+- Zk-proof of the state transitions, as traditional zk-rollups do
+- Zk-proof of the rollup data to the data root tuple root
+- Verify that the sequence of spans is valid, i.e., is part of the Celestia block referenced by its height.
+
+Once these are valid, the settlement contract can be sure that the state transitions are valid, and the data was posted to Celestia.
+
+Asking inputs from reviewers: when verifying that the sequence of spans exists, and the rollup data to the data root tuple root, how would we know that we're talking about the same data? I could create a zk-proof of a bunch of data to the data root tuple root, then provide an invalid sequence of spans, how would this be detected?
```
rach-id, line 54: This is for reviewers :D

jcstein:
I'll review and add to menu when the time comes

developers/blobstream-rollups.md
```diff
@@ -0,0 +1,123 @@
...
+---
+
+## Optimistic rollups
+
+One type of rollups that can be built with Blobstream is optimistic rollups.
+An optimistic rollup is a rollup that commits optimistically to a set of blocks,
+and allows the other parties to verify that the blocks are valid,
+and if they're not, they can create fraud proofs to signal that.
+
+Through the use of Blobstream, Celestia allows optimistic rollups to use Celestia as a DA layer,
```
evan-forbes, line 12: [opinion]

as far as intros, I think leading with this is best, if we're aiming at rollup devs here, then I think this would be better than prefacing with what an optimistic rollup is. We can could just say this sentence, but then we don't even have to mention optimistic rollups, just optimistic rollups in general.

if the goal of this post is to explain how to create commitments to rollups, then we should lead with that.

I think this intro also assumes that the reader is aware that the commitments are done differently, however the purpose of this doc appears to be to explain that there is a difference and this is a good starting point.

developers/blobstream-rollups.md
```diff
@@ -0,0 +1,123 @@
...
+i.e., posting all the rollup data to Celestia, and only send commitments to the settlement contract.
+Then,
+users can use Blobstream as a source of truth to verify the rollup data
+and create fraud proofs in case of a misbehavior.
+
+To build an optimistic rollup that uses Celestia as a DA layer, the following constructions can be inspired by.
+
+### Optimistic rollups that uses a sequence of spans
+
+One way to construct an optimistic rollup that uses Celestia as a DA layer is to post the rollup data in Celestia.
```
evan-forbes, line 22: I found this sentence a bit confusing, don't all celestia rollups have to post data to celestia? is there another way?

developers/blobstream-rollups.md
```diff
@@ -0,0 +1,123 @@
...
+Then in the rollup header, use the following information to reference a sequence of spans, aka a data pointer:
```
evan-forbes, line 23: in the intro, I would first go into the details of a span if its used for both. Below we reference this span section, but its non-intuitive, since zk rollup devs aren't going to read the optimistic rollup section

developers/blobstream-rollups.md
```diff
@@ -0,0 +1,123 @@
...
+
+- `Height`: the height of the Celestia block containing the rollup data
+- `Start share`: the index of the first share containing the rollup data
+- `Data length`: the number of shares that the rollup data occupies
```
evan-forbes, line 27: These are some of the key things to understand before anything else. imo, we should first describe Spans in their own section

if we already have span docs, then we can just point to those, but I think the important bit is have a section that people can reference when they forget what a span is.

developers/blobstream-rollups.md
```diff
@@ -0,0 +1,123 @@
...
+
+Note:
+If the rollup data is submitted in multiple blocks,
+the above sequence of spans can be generalized to include multiple blocks.
+For simplicity, we will stick with the data only submitted to a single Celestia block.
+
+Now that this information is saved in the rollup header/settlement contract,
+and users/rollup full nodes will be able to query for the rollup data from Celestia and verify that it's valid.
+And if it's not, fraud proofs can be generated.
```
evan-forbes, line 36: this is great info! and I think should be part of the span section

something like "Spans reference aribtrary amounts of data across multiple blocks" this automatically tells me what a span is at a high level

developers/blobstream-rollups.md
```diff
@@ -0,0 +1,123 @@
...
+
+The proofs in this setup are discussed in the [inclusion proofs](https://github.com/celestiaorg/blobstream-contracts/blob/master/docs/inclusion-proofs.md) documentation.
+
+### Optimistic rollups that use share commitments
+
+Another way to build a rollup is to replace the sequence of spans with a `height` and a `share commitment`.
```
evan-forbes, line 42: when talking about spans or sequences in general, I think we should mention their different functionalities without actually talking about rollups first, then we tie in other sections how spans can be used for different rollup types

developers/blobstream-rollups.md
```diff
@@ -0,0 +1,123 @@
...
+  - share merkle proof up to the [subtree root](https://celestiaorg.github.io/celestia-app/specs/data_square_layout.html#blob-share-commitment-rules)
+    corresponding to that share
+  - subtree root merkle proof to the [share commitment](https://celestiaorg.github.io/celestia-app/specs/data_square_layout.html#blob-share-commitment-rules)
+- share commitment inclusion to the data root tuple root: meaning four merkle proofs:
+  - subtree roots merkle proofs to the share commitment: to make sure the subtree roots are valid
+  - subtree roots merkle proofs up to the row roots: to prove that the subtree roots belong to a set of rows
+  - row roots proofs to the data root
+  - data root tuple proof to the data root tuple
+
+Once these are valid, the rollup contract can proceed to parse the share and verify the contested claim.
```
evan-forbes, line 63: "contested claim" is an undefined phrase, I'm not sure what that means. Is that a fraud proof?

developers/blobstream-rollups.md
```diff
@@ -0,0 +1,123 @@
...
+
+Another alternative when using the share commitment to reference the rollup data is to parse the PFB,
+i.e., the protobuf serialized PayForBlob transaction,
+inside the rollup settlement contract and verify that the parsed share commitment is the same.
+Then, proving the inclusion of the PFB to the data root tuple root.
+More details on this can be found in [ADR-011](https://github.com/celestiaorg/celestia-app/blob/main/docs/architecture/adr-011-optimistic-blob-size-independent-inclusion-proofs-and-pfb-fraud-proofs.md). 
+
+The issue with this approach is needing a library to parse the protobuf in solidity and the gas cost that comes with it.
+Otherwise, proving the PFB to the data root tuple root is cheaper than proving the blob,
+because it needs only proving two shares.
```
evan-forbes: this separate section is good! I think we should try to do this with all of the ways to prove data. Agnostic to how rollups actually use them. Most rollup devs know how their proving systems work, so if we just tell them about the different ways to prove data, then they can connect that to their proving system.

we can ofc discuss our thoughts on how these things work in a zk or optimistic rollup later as well

developers/blobstream-rollups.md
```diff
@@ -0,0 +1,123 @@
...
+
+## Zk-Rollups
+
+Zk-rollups, aka validity rollups, can also use Celestia as a DA and Blobstream to verify that the data was posted.
+However, the submission process is different from the above constructions,
```
evan-forbes, line 86: after listing/describing the different ways to prove data, then I think that we can use a sentence similar to 

> However, the submission process is different from the above constructions

something along these lines, where can start the rollup section with "How data inclusion is proved is highly dependant upon the rollup's proving system. Then discuss ORU, then ZKR

developers/blobstream-rollups.md
```diff
@@ -0,0 +1,123 @@
...
+
+So, when posting to the settlement contract,
+the rollup data in Celestia can be referenced either using a `share commitment` or a `sequence of spans`,
+and a `height`, similar to the previous constructions.
+
+### Zk-rollups that uses a sequence of spans
+
+So, when submitting the headers to the rollup settlement contract,
+the settlement contract will need to verify the following:
+
```
evan-forbes, line 97: as mentioned elsewhere, this is good to describe! I think we should first define a sequence of spans, then explain the tradeoffs for a zk rollup using this system, or what it should likely do

developers/blobstream-rollups.md
```diff
@@ -0,0 +1,123 @@
...
+- Zk-proof of the rollup data to the data root.
+  The verification process of this should accept a commitment as input
+  so that the contract makes sure it's the correct value that's being saved.
+  The commitment can be the data root and the sequence of spans.
+  And, when the rollup data is proven inside the circuit to the data root,
+  the used data root is asserted to be the input one.
+  Similarly, the data's location is asserted to be the same as the input sequence of spans.
+  While these same arguments are the ones used as part of the `sequence of spans` verification above.
+
+Once these are valid, the settlement contract can be sure that the state transitions are valid,
```
evan-forbes, line 111: in general, I think this post tip toes into rollup proving systems, but imo, this actually distracts from the main points that we are trying to get at.

evan-forbes (commented):
really good first draft! I think this could be super helpful to devs in actually understanding how this works, and could be the thing that we point rollup devs to in the future

imo, I do think we can clarify a bit of the content, and reorganize it a bit. again, imo, a potential structure could be 

- very quick intro into blobstream
- explain each mechanism to prove inclusion as precisely, concisely, and isolated as we can
- then briefly mention that zk and oru s differ in how they use blobstream. the optimistic vs pessimistic nature
- how an ORU would use each described mechanism for proving inclusion. if you don't think it makes sense to use this construction explain why! this would be super informative
- how a zkr would use this, what challenges does the pessimistic approach run into?





one general thing about this post is that it assumes that the reader is reading the entire things from start to finish, but I don't think this usally occurs with technical docs. If we can, I think it would add a bit of clarity if we can order things a bit differently

rach-id:
@evan-forbes Thanks a lot for the feedback. I rewrote the whole document to be almost self-contained. Please review when you have time

developers/blobstream-rollups.md
```diff
@@ -0,0 +1,399 @@
...
+## Concepts
+
+### Share commitment
+
+The [share commitment](https://celestiaorg.github.io/celestia-app/specs/data_square_layout.html#blob-share-commitment-rules) is a commitment over the data
+contained in the [PayForBlob](https://github.com/celestiaorg/celestia-app/blob/v1.0.0-rc2/proto/celestia/blob/v1/tx.proto#L16-L31) transaction.
+This commitment allows
+proving that the corresponding data exists on Celestia [efficiently](https://github.com/celestiaorg/celestia-app/blob/main/docs/architecture/adr-011-optimistic-blob-size-independent-inclusion-proofs-and-pfb-fraud-proofs.md).
+
+#### Proof details
```
evan-forbes, line 24: do we really have 6 different proof types? if so, do users need to know about all 6? in this case, perhaps we should only focus on the proofs that the users need (the proofs returned by the api), then describe what each consists of if needed

developers/blobstream-rollups.md
```diff
@@ -0,0 +1,399 @@
...
+
+To prove that the data corresponding to a share commitment was posted to Celestia using Blobstream,
+the following proofs need to be verified:
+
+- share inclusion proof to the share commitment: meaning creating two merkle proofs:
+  - share merkle proof up to the [subtree root](https://celestiaorg.github.io/celestia-app/specs/data_square_layout.html#blob-share-commitment-rules)
+    corresponding to that share
+  - subtree root merkle proof to the [share commitment](https://celestiaorg.github.io/celestia-app/specs/data_square_layout.html#blob-share-commitment-rules)
+- share commitment inclusion proof to the data root tuple root: meaning four merkle proofs:
```
evan-forbes, line 33: when would users of blobstream create one of these again? do we know if this is less efficient than just using a share proof to the data commitment

developers/blobstream-rollups.md
```diff
@@ -0,0 +1,399 @@
...
+An example optimistic rollup
+that uses sequence of spans
+to reference its data can be found in the [`RollupInclusionProofs`](https://github.com/celestiaorg/blobstream-contracts/blob/master/src/lib/verifier/test/RollupInclusionProofs.t.sol).
+It portrays the different possible data availability proofs, constructs them and shows how to verify them.
+
+Also,
+more details
+on querying these kinds of proofs can be found in the [proof queries](https://docs.celestia.org/developers/blobstream-proof-queries) documentation.
+
+### Optimistic rollups that use share commitments
```
evan-forbes, line 257: if there really are no cons the first scheme, will anyone use this scheme? if so, why? maybe list that as a con above. If not, then perhaps we can remove this section

developers/blobstream-rollups.md
```diff
@@ -0,0 +1,399 @@
...
+
+#### Pros
+
+- The inclusion proof inside the zk-circuit is a simple proof that uses traditional merkle tree.
+  In the case of using share commitment, as will be explained below,
+  additional libraries that can be expensive to prove are required.
+
+#### Cons
+
+None
```
evan-forbes, line 328: same question as before, if there are no cons, then can we remove other sections? if not, then we should add cons here

evan-forbes (commented):
super nice!

developers/blobstream-rollups.md
rach-id: The reason I didn't include API is that we still don't have it... So I decided to go low level to explain the proofs. One other problem is that we don't have any document focusing on explaining the square layout in a beginner friendly way...

developers/blobstream-rollups.md
rach-id: yes, it's less efficient because you need an extra proof of the share to the share commitment

developers/blobstream-rollups.md
```diff
@@ -0,0 +1,399 @@
...
+  this requires
+  waiting for the Celestia block to be committed to by Blobstream before saving updating the settlement contract.
+  This might require waiting for a few hours, depending on the batches size on each chain,
+  to finally submit the rollup update.
+
+Given these limitations, an alternative design will be discussed in the next section.
+
+### Sequence of spans
+
+An alternative, yet preferred,
```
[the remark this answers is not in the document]
rach-id: removing it 👍

developers/blobstream-rollups.md
rach-id: I want to include this construction because it's how current rollups think it should be done, then add cons for it

developers/blobstream-rollups.md
evan-forbes: thanks for clarifying! that makes sense and its okay to wait, but I do think we should try to simply things as much as possible given the proofs are so layered

developers/blobstream-rollups.md
```diff
@@ -0,0 +1,399 @@
...
+
+To prove that the data corresponding to a share commitment was posted to Celestia using Blobstream,
+the following proofs need to be verified:
+
+- share inclusion proof to the share commitment: meaning creating two merkle proofs:
+  - share merkle proof up to the [subtree root](https://celestiaorg.github.io/celestia-app/specs/data_square_layout.html#blob-share-commitment-rules)
+    corresponding to that share
+  - subtree root merkle proof to the [share commitment](https://celestiaorg.github.io/celestia-app/specs/data_square_layout.html#blob-share-commitment-rules)
+- share commitment inclusion proof to the data root tuple root: meaning four merkle proofs:
+  - subtree roots merkle proofs to the share commitment: to make sure the subtree roots are valid
```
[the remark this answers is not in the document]
rach-id: Once https://github.com/celestiaorg/celestia-app/pull/3245 is merged, we can link to it in these proofs to reduce confusion

developers/blobstream-rollups.md
rach-id: Once https://github.com/celestiaorg/celestia-app/pull/3245 is merged, we can link to it in these proofs to reduce confusion

developers/blobstream-rollups.md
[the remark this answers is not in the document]
rach-id: [0545d7f](https://github.com/celestiaorg/docs/pull/1489/commits/0545d7fba3f5ebd49c963507a25da8f9f0263acc) Adds links to get more details what each proof means

developers/blobstream-rollups.md
rach-id: [0545d7f](https://github.com/celestiaorg/docs/pull/1489/commits/0545d7fba3f5ebd49c963507a25da8f9f0263acc) adds information on these proofs

evan-forbes (approved):
I have a some comments, but think they can be addressed later and that we shouldn't block on this any longer

great work @rach-id and thanks for pushing this through

jcstein:
schitts-creek-what-gif-by-cbc
pardon my rebase

jcstein:
@rach-id in https://github.com/celestiaorg/docs/pull/1489/commits/89b78bf54557aac073aafff6b2a2432b2116e001 I removed duplicate headings for readability/linkability

in https://github.com/celestiaorg/docs/pull/1489/commits/bf2bf7bf4c0a28a093b5f0e269b64448f6e5879e I clarified the duplicate headings. can you confirm that in https://github.com/celestiaorg/docs/pull/1489/commits/bf2bf7bf4c0a28a093b5f0e269b64448f6e5879e I named the links to headings correctly?

other than that, this LGTM.

rach-id:
@jcstein why did we add the headings? I thought that if we have the heading in `###` and the subsections in `####`, it will show that it's part of it no? 
If not, then I can rewrite the headings to have more details

jcstein:
to avoid breaking linting rules. i think that they would exist as "#pros", "#pros-1" etc. but are more clear with specific headings to the section they're relevant to

however, if you'd prefer, i can revert the changes and ignore the linting rules

rach-id:
@jcstein fixed the headings, should be good to go

jcstein:
gm @rach-id it is under "Developers > Build whatever (formerly rollup category)"
would you like to move it into blobstream instead?

developers/blobstream-rollups.md
```diff
@@ -0,0 +1,452 @@
...
+of users. It allows rollups to post their data on Celestia while proving their
+availability in the rollup settlement contract.
+
+This document will outline a few ways to build optimistic or zk-rollups
+that post their data to Celestia and use Blobstream to prove that data's
+availability.
+
+## Concepts
+
+### Share commitment
```
adlerjohn, line 18: Two issues here:
1. It's not clear why "share commitment" is used here instead of "blob commitment" as used everywhere else.
2. For the reader reading from top to bottom, it's not obvious until the next subsection that this is only one way of doing things, rather than the only way.

rach-id: 1. I saw it in the specs being called `blob share commitment` that's why I used `share commitment`. Will change ASAP.

2. good catch, will mention that

rach-id: @adlerjohn check here: https://github.com/celestiaorg/docs/pull/1541

PR #1557, opened 2024-05-02
developers: clarify why grpc endpoint needed for light node

renaynay:
## Summary by CodeRabbit

- **Documentation**
	- Enhanced the tutorial on connecting to a public core endpoint for the Celestia Light node, emphasizing access to state queries, transactions, and other state-related queries.
	- Added information about connecting to a public core endpoint using `--core.ip string` in various node tutorials to provide access to state queries and transactions.

### DIFF
--- a/developers/node-tutorial.md
+++ b/developers/node-tutorial.md
@@ -305,9 +305,9 @@ your keys will be stored.
 ### Connect to a public core endpoint
 
 Let's now run the Celestia Light node with a gRPC connection
-to an example public core endpoint. Connecting to a public core endpoint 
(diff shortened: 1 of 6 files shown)


jcstein:
thank you @renaynay - I've added some redundancy for this explanation. one thought, should we just call it a "core endpoint" instead of a "public core endpoint"?

jcstein (approved):
I can add back "public" if you think it is necessary. I am hoping it may reduce confusion for users and make it more clear they can use any "core endpoint", not just public ones

PR #1564, opened 2024-05-07
docs: Increase LN disk recommendation to 100GB and bridge/full RAM to 16GB

Wondertan:
It's supposed to be temporary to give us some buffer until the sample and header pruning is finished. Afterward, we can revert and even decrease to ~5GB with pruning milestones finished.

## Summary by CodeRabbit

- **Documentation**
    - Updated the minimum required SSD storage for a light node from 50 GB to 100 GB.
    - Updated the minimum recommended memory requirement for bridge and full-storage nodes to 16 GB RAM.


jcstein:
TODO
- [ ] revise diagram to say 100GB

jcstein (approved):

PR #1568, opened 2024-05-10
docs: add note about slashing

jcstein:
## Summary by CodeRabbit

- **Documentation**
	- Added a warning in the documentation about Blobstream X being in beta and that slashing is not yet enabled.


rach-id (changes requested):
requesting changes so we don't merge by accident

rach-id (approved):

PR #1605, opened 2024-06-11
Update foundation-delegation-program.md

mindstyle85:
## Summary by CodeRabbit

- **Documentation**
  - Updated the foundation delegation program guidelines to include a new requirement for participants to maintain a fully archival bridge node for both Mainnet Beta and Mocha.


mindstyle85:
adding what we discussed on the call to the docs here @YazzyYaz , lmk if its ok, ive written it twice just so people really make sure theyre aware of it. We can keep this PR open until this is final and we are sure it wont change

jcstein (approved):

PR #1608, opened 2024-06-14
feat: add section on maxblob size

jcstein:
## Overview

[PREVIEW](https://celestiaorg.github.io/docs-preview/pr-1608/developers/submit-data#maximum-blob-size)
Screenshot 2024-06-14 at 3 51 26 PM

- [x] Adds a section on max blob size to mainnet beta page https://github.com/celestiaorg/docs/commit/4f4a3cf7aa5fe15f73878a28f6a21b2edd607383
- [x] a link to the long description from the short description https://github.com/celestiaorg/docs/commit/4f4a3cf7aa5fe15f73878a28f6a21b2edd607383
- [x] adds link on node-tutorial in callout to the new, shorter description https://github.com/celestiaorg/docs/pull/1608/commits/5f8768bc2c724fc6cf29cb4d65b5fd2730d0ad28
- [x] adds a link to code in the long description https://github.com/celestiaorg/docs/pull/1608/commits/5044134210a2e268407d514c0f6f60e4d05b6b45

## Summary by CodeRabbit

- **Documentation**
  - Updated guidelines on maximum blob size in a transaction.
  - Added a tip on maximum blob size in the Node tutorial.
  - Included references to code snippets for blob size calculations in the mainnet documentation.

### DIFF
--- a/developers/submit-data.md
+++ b/developers/submit-data.md
@@ -12,6 +12,16 @@ To submit data to Celestia, users submit blob transactions (`BlobTx`). Blob
 transactions contain two components, a standard Cosmos-SDK transaction called
 `MsgPayForBlobs` and one or more `Blob`s of data.
 
+## Maximum blob size
+
+The maximum total blob size in a transaction is just under **2 MiB
+(1,973,786 bytes)**, based on a 64x64 share grid (4096 shares).
+With one share for the PFB transaction, 4095 shares remain:
+1 at 478 bytes and 4094 at 482 bytes each.
+
+This is subject to change based on governance parameters.
+Learn more on [the Mainnet Beta page under "Maximum bytes"](../nodes/mainnet.md#maximum-bytes).
+
 ## Fee market and mempool
 
 Celestia makes use of a standard gas-priced prioritized mempool. By default,
--- a/developers/submit-data.md
+++ b/developers/submit-data.md
@@ -22,6 +22,12 @@ With one share for the PFB transaction, 4095 shares remain:
 This is subject to change based on governance parameters.
 Learn more on [the Mainnet Beta page under "Maximum bytes"](../nodes/mainnet.md#maximum-bytes).
 
+It is advisable to submit transactions where the total blob size is significantly
+smaller than 1.8 MiB (e.g. 500 KiB) in order for your transaction to get included
+in a block quickly. If a tx contains blobs approaching 1.8 MiB then there will
+be no room for any other transactions. This means that your transaction will
+only be included in a block if it has a higher gas price than every other
+transaction in the mempool.
 ## Fee market and mempool
 
 Celestia makes use of a standard gas-priced prioritized mempool. By default,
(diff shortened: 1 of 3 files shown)


developers/submit-data.md
```diff
@@ -12,6 +12,16 @@ To submit data to Celestia, users submit blob transactions (`BlobTx`). Blob
 transactions contain two components, a standard Cosmos-SDK transaction called
 `MsgPayForBlobs` and one or more `Blob`s of data.
 
+## Maximum blob size
+
+The maximum total blob size in a transaction is just under **2 MiB
```
jcstein, line 17: Although this is technically the max blob size, I think we should provide an advisable maximum size of something like 80-90% of the max size to avoid timeouts on users submitting large blobs.

distractedm1nd: This doesn't make sense to me yet WHY it happens. I know things would get congested if the entire block gets filled, but I wasn't aware it happens to individual blobs as well?

jcstein: if the whole block is filled with just 1 blob then there is no room for any other transactions, so it will most likely take longer to get it included, unless you really jack up the gas

jcstein: so i think encouraging people to not try to push the limit would help things get included without wasting gas on beating the lower fees out

distractedm1nd (approved):
Thank you for explaining, makes sense

developers/submit-data.md
```diff
@@ -12,6 +12,16 @@ To submit data to Celestia, users submit blob transactions (`BlobTx`). Blob
...
+(1,973,786 bytes)**, based on a 64x64 share grid (4096 shares).
+With one share for the PFB transaction, 4095 shares remain:
+1 at 478 bytes and 4094 at 482 bytes each.
+
+This is subject to change based on governance parameters.
+Learn more on [the Mainnet Beta page under "Maximum bytes"](../nodes/mainnet.md#maximum-bytes).
+
```
jcstein, line 24:
```suggestion

It is advisable to _not_ try to fill the entire block when submitting blobs,
as a block filled with just one large blob can delay transaction inclusion. Avoiding
blobs that take up the maximum block capacity also ensures smoother transactions
and reduces unnecessary gas expenditures.
```

jcstein: need to iron this out..

rootulp: Agreed with your points on the other thread. Perhaps they can be included here:

It is advisable to submit transactions where the total blob size is significantly smaller than 1.8 MiB (e.g. 500 KiB) in order for your transaction to get included in a block quickly. If a tx contains blobs approaching 1.8 MiB then there will be no room for any other transactions. This means that your transaction will only be included in a block if it has a higher gas price than every other transaction in the mempool.

rootulp (approved):

developers/submit-data.md
jcstein: used that, thank you! https://github.com/celestiaorg/docs/pull/1608/commits/520f679498883774b8a349990807e6899695d5cb

PR #1615, opened 2024-06-18
docs: replace gem image

qubelabsio:
## Overview

Added correct dimensions to match with others.


jcstein:
good call and ty @qubelabsio

jcstein (approved):

PR #1617, opened 2024-06-18
fix: setting node doc links to use mainnetVersions

distractedm1nd:
Self-explanatory

## Summary by CodeRabbit

- **Documentation**
  - Updated URLs in multiple developer guides to dynamically reference the latest version of the Celestia node documentation.
  - Improved accuracy and flexibility by ensuring API endpoint links always point to the latest version.
  - Ensured users access the most up-to-date documentation for methods like `blob.Submit`, `blob.GetAll`, `header.Subscribe`, and `share.GetEDS`.

### DIFF
--- a/developers/blobstream-offchain.md
+++ b/developers/blobstream-offchain.md
@@ -2,6 +2,10 @@
 description: Learn how to integrate your L2's offchain logic with Blobstream
 ---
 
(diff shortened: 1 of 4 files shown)


jcstein (approved):
good call. thank you

PR #1674, opened 2024-08-20
docs: add updated diagrams for node requirements and consensus node

jcstein:
## Overview
1. Updates diagram for node requirements
    1. Q: Should we take out the plain text version? I was thinking it could be useful for SEO purposes. But the duplication feels odd.
1. Updates consensus node diagram to say "consensus node" instead of "full consensus node"

## Summary by CodeRabbit

- **Documentation**
	- Updated the image format in the consensus node documentation for improved clarity.
	- Reintroduced a visual banner in the "Recommended Celestia node requirements" section to enhance user interface.
	- Removed a commented-out image banner to improve the visual presentation.


cmwaters (approved):
Do consensus nodes really need 2TB? Do we make a distinction between pruned nodes and archival nodes

rootulp (approved):
> Q: Should we take out the plain text version? I was thinking it could be useful for SEO purposes. But the duplication feels odd.

The duplication feels like a maintenance burden. If we had to take out one of them, my preference would be to drop the image because the plaintext version is easier to edit. 

Given the image was already created, I don't think we need to drop it for now. Ultimately defer to you.

jcstein:
I'm also leaning toward removing the maintenance burden of this graphic, too, and keeping the plaintext version. Will revisit in an hour or so... Thanks for the feedback!

PR #1816, opened 2024-11-28
Update celestia-node.sh to use curl instead of wget

lukecd:
This update modifies celestia-node.sh to use curl instead of wget as curl is installed by default on new macs

## Summary by CodeRabbit

- **New Features**
	- Enhanced download mechanism using `curl` for improved reliability.
- **Bug Fixes**
	- Improved error handling during the download process to ensure failures are logged and handled appropriately.


jcstein (approved):
Resolves #1816

PR #1944, opened 2025-03-21
docs: added shields on stars, forks and links to them

sashaphmn:


jcstein (approved):
OK

jcstein (changes requested):
actually, this is a good idea but this info is already on the page where user would see readme. i think it's best we leave these out as it will get too cluttered

jcstein:
there's a codeql workflow here??

jcstein:
sure

sashaphmn:
Of course, the final decision is up to you, but two icons, Goland-Linter and Test, have been added to the celestia-core repository for workflows.

PR #2012, opened 2025-04-28
doc: improve the page of setting up a validator node

gluckzhang:
## Overview

This PR contains the following two improvements for the page [Setting up a Celestia validator node](https://docs.celestia.org/how-to-guides/validator-node).

First, the sub-section "Delegate stake to a validator" is moved a bit later after creating a validator. If this is done before creating the validator, the transaction will fail with an error "failed to execute message; message index: 0: validator does not exist" (example [here](https://testnet.celestia.explorers.guru/transaction/CEDAC5D415A774E37CA4F4C5D04E6D0EE4F532A9D66038BC5C64ADEC44559681?height=5857805)).

Second, when creating a validator, users can add extra information such as the logo, website, and a description for the validator. It is also possible to update these fields after creating a validator. This PR adds such explanations to the related section.

## Summary by CodeRabbit

- **Documentation**
  - Improved the validator node setup guide by reorganizing and clarifying delegation instructions.
  - Clarified usage of the validator wallet variable.
  - Enhanced validator creation steps with optional metadata fields and added guidance on updating validator details after creation.

### DIFF
--- a/how-to-guides/validator-node.md
+++ b/how-to-guides/validator-node.md
@@ -30,60 +30,6 @@ First, follow the instructions on
 
 Follow [the tutorial on creating a wallet](/how-to-guides/celestia-app-wallet.md).
 
-### Delegate stake to a validator
-
-Create an environment variable for the address:
-
-```bash
-VALIDATOR_WALLET=<validator-wallet-name>
-```
-
-If you want to delegate more stake to any validator, including your own you
-will need the `celestiavaloper` address of the validator in question. You can run
-the command below to get the `celestiavaloper` of your local validator wallet in
-case you want to delegate more to it:
-
-```bash
-celestia-appd keys show $VALIDATOR_WALLET --bech val -a
-```
-
-After entering the wallet passphrase you should see a similar output:
-
-```bash
-Enter keyring passphrase:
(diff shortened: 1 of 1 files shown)


how-to-guides/validator-node.md
```diff
@@ -234,6 +202,60 @@ You should now be able to see your validator from
 After starting your node, please submit your node as a seed and peer to the
 [networks repository](https://github.com/celestiaorg/networks).
 
+## Optional: Delegate more stake to a validator
```
jcstein, line 205: Thank you for the PR @gluckzhang. why does this heading change from "### Delegate stake to a validator" to "## Optional: Delegate more stake to a validator" i.e. why is it "More" now?

gluckzhang: Hi @jcstein, thanks for reviewing it :) Because I moved the section later and I think it should be at the same level as another optional step. Maybe we can group some of the sub-sections if you think the page contains too many `##` headlines. Just let me know what you think and I will adjust it!

Here are the screenshots that show the difference.

Current proc version:
image

This PR:
image

gluckzhang: Oh sorry I just got your question. I agree `more` is not necessary here. Let me change it back to the original title :)

jcstein (approved):
thanks again!

PR #2032, opened 2025-05-13
feat: add llms.txt workflow

jcstein:

### DIFF
--- a/.github/workflows/generate_llms_txt.yml
+++ b/.github/workflows/generate_llms_txt.yml
@@ -1,0 +1,30 @@
+name: Generate LLMs.txt File
+
+on:
+  # Run on push to main branch
+  push:
+    branches: [main]
+  # Allow manual triggering
+  workflow_dispatch:
+
+jobs:
+  generate:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        
+      - name: Setup Node
+        uses: actions/setup-node@v4
+        with:
+          node-version: 18
+          
+      - name: Generate llms.txt
+        run: npx -y llmstxt gen https://docs.celestia.org/sitemap.xml > public/llms.txt
+        
+      - name: Commit and push if changed
+        run: |
+          git config --global user.name 'GitHub Actions'
+          git config --global user.email 'actions@github.com'
+          git add public/llms.txt
+          git diff --quiet && git diff --staged --quiet || (git commit -m "Update llms.txt file" && git push)


.github/workflows/generate_llms_txt.yml
```diff
@@ -0,0 +1,30 @@
...
+      - name: Checkout
+        uses: actions/checkout@v4
+        
+      - name: Setup Node
+        uses: actions/setup-node@v4
+        with:
+          node-version: 18
+          
+      - name: Generate llms.txt
+        run: npx -y llmstxt gen https://docs.celestia.org/sitemap.xml > public/llms.txt
```
jcstein, line 23: I realized this is using the site to build it, so wouldn't build it from the new content being changed/added. so maybe this should be a cron job 🤔

jcstein: https://github.com/celestiaorg/docs/pull/2032/commits/ae4af0ebf9bd487aed9244e3c964f241229006ff

gbarros: I think post-merge would be even better:

image

```yaml
on:
  push:
    branches:
      - main  # or 'master', whatever your default branch is
    paths:
      - docs/**       # adjust these patterns to only match relevant changes
      - .github/workflows/generate_llms_txt.yml

```

.github/workflows/generate_llms_txt.yml
```diff
@@ -0,0 +1,30 @@
...
+        
+      - name: Commit and push if changed
+        run: |
+          git config --global user.name 'GitHub Actions'
+          git config --global user.email 'actions@github.com'
+          git add public/llms.txt
+          git diff --quiet && git diff --staged --quiet || (git commit -m "Update llms.txt file" && git push)
```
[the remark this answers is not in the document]
gbarros: This also looks ok, do you want to commit it? @jcstein

.github/workflows/generate_llms_txt.yml
jcstein: i think `/docs` doesn't exist. would the post-merge work still if the site hasn't built, since it's generating it from the built site?

gbarros: I hadn't noticed/focused on it. I think you are correct. 
In this case, removing this "paths" key altogether should be fine, really, because the cmd is reading from the deployed website.

gbarros: Also, I don't know how deployment is working for now, but I am guessing it takes a moment for it to be live after merge, right? 

So I got 2 decent solutions to ponder: 
image
or 

image

jcstein: let's do sleep for 10 mins then build it, should only take 5 mins max to build but better to be on safe side

PR #2043, opened 2025-05-20
Update celestia-app-metrics.md with Tenderduty

bernatfp:
Add reference to [Tenderduty](https://github.com/Firstset/tenderduty) as an alternative for monitoring and alerting.

We use both Grafana and Tenderduty to monitor our Celestia testnet validator, as well as our other Tendermint validators.

We think it may be valuable for fellow operators to be aware of it.


jcstein:
Is this advisable @mindstyle85 ?

PR #2052, opened 2025-05-29
Update block timing in docs

sezna:
## Overview

I noticed the docs refer to the older block timing in this section. Now that blocks are every 6 seconds, 5 blocks is roughly 30 seconds, not a minute.

## Summary by CodeRabbit

- **Documentation**
  - Updated the estimated time for nodes to drop transactions from the mempool, correcting it from roughly 1 minute to roughly 30 seconds after 5 blocks.


jcstein (approved):
LGTM, thank you!

PR #2097, opened 2025-06-05
Clarify TxConfig initialization in golang-client-tutorial.md

thearcyx:
## Overview

This PR improves the documentation of the `NewTxConfig` function in `golang-client-tutorial.md`.

The updated comment explains that `NewTxConfig()` initializes a transaction configuration with default gas settings (`DefaultGasPrice`, `DefaultMaxGasPrice`) and that it can optionally take variadic options to customize its behavior.

This change is important because passing `nil` as the `TxConfig` can lead to runtime errors. For example, when calling `client.Blob.Submit(ctx, []*blob.Blob{...}, nil)`, although `nil` is accepted by the function signature, it may cause internal panics due to uninitialized fields in the transaction context. Using a default `TxConfig` avoids this and ensures stability.

## Context

While following the Go client tutorial, I encountered a runtime error when submitting a blob using `nil` for the `SubmitOptions`. This was resolved by explicitly initializing a `TxConfig` using `state.NewTxConfig()`. This PR updates the guide to help future developers avoid the same pitfall.

## Checklist

- [x] Clear and safe example added for TxConfig usage
- [x] Based on a reproducible runtime issue
- [x] Aligned with Celestia's contributor guide
- [x] No breaking changes introduced

## Summary by CodeRabbit

- **Documentation**
  - Updated the Golang client tutorial to use a default transaction configuration for blob submission, ensuring safer default settings.
  - Added comments to clarify the use of a default transaction configuration.

### DIFF
--- a/tutorials/golang-client-tutorial.md
+++ b/tutorials/golang-client-tutorial.md
@@ -57,6 +57,7 @@ import (
     client "github.com/celestiaorg/celestia-node/api/rpc/client"
     "github.com/celestiaorg/celestia-node/blob"
     share "github.com/celestiaorg/go-square/v2/share"
+    "github.com/celestiaorg/celestia-node/state"
 )
 
 // SubmitBlob submits a blob containing "Hello, World!" to the 0xDEADBEEF namespace. It uses the default signer on the running node.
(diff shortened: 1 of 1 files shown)


jcstein:
let's also update the CI/CD that was added in https://github.com/celestiaorg/docs/pull/2069 for this

jcstein:
Closing as obsolete.

The target file (`tutorials/golang-client-tutorial.md`) was removed in the VitePress → Nextra migration. The current Go client tutorial (`app/build/post-retrieve-blob/client/go/page.mdx`) uses `api/client`, which intentionally accepts `nil` for submit options (and upstream examples do the same). The related CI follow-up from #2069 is also outdated after that stack rewrite.

Thanks for the contribution!

PR #2162, opened 2025-07-22
docs: mainnet upgrade height for lotus

jcstein:


jcstein:
did this while in a call before seeing #2161

PR #2203, opened 2025-08-12
chore: remove go.mod

rootulp:
My editor was complaining because a go mod tidy was needed in this repo.

TBH I'm not sure what this go.mod is for so if it's unused then we could delete instead of merging this PR.


jcstein:
thank you for the PR. it's an a(i)ttempt at a CI test for example go blob submission: https://github.com/celestiaorg/docs/pull/2069

i think it could honestly be removed

rootulp:
Sounds good, removed go.mod.

jcstein (approved):

PR #2274, opened 2025-10-24
docs: add light node guide

jcstein:
## Overview

WIP, need to

- [ ] add variables to docs so we can use network versions in multiple places
- [x] figure out if we need this guide to have trusted hash so people dont wait on the node to sync...after 30m mine still can't retrieve the blob i posted. nobody who wants to do a quickstart will just wait for an hour before getting to the next step. but setting trusted hash shouldn't be required..


jcstein:
while adding constants i'm remembering this can get messy..

if using in code block, doesn't render:

Screenshot 2025-10-27 at 10 43 40

if using custom code block, well, you have to make a custom code block... it breaks the UI

Screenshot 2025-10-27 at 10 41 59

PR #2288, opened 2025-10-29
docs: update tx client guide to rm unecessary port

jcstein:
the port isn't necessary, cc @Ferret-san


jcstein:
disregard bc that breaks with `failed to get node info: rpc error: code = Unavailable desc = name resolver error: produced zero addresses`

PR #2290, opened 2025-10-29
[blocked] change pull request trigger to use pull_request and update check…

gbarros:
This pull request updates the GitHub Actions workflow for PR previews by changing the trigger event and the way the PR code is checked out. These changes help ensure that previews are built using the correct commit and improve security by using the standard `pull_request` event.

**Workflow trigger changes:**

* The workflow now uses the `pull_request` event instead of `pull_request_target`, which is safer and more appropriate for building PR previews since it runs in the context of the PR branch and avoids potential security risks.

**Checkout logic changes:**

* The checkout step now uses `github.sha` instead of `github.event.pull_request.head.sha`, ensuring the workflow checks out the exact commit that triggered the workflow, which is more reliable and consistent for preview builds.

### DIFF
--- a/app/build/post-retrieve-blob/client/page.mdx
+++ b/app/build/post-retrieve-blob/client/page.mdx
@@ -17,7 +17,11 @@ The Celestia Go client lets you submit and retrieve data from the Celestia netwo
 
 ## Quick setup
 
-### 1. Create your project
+import { Steps } from 'nextra/components'
+
+<Steps>
+
+### Create your project
 
 ```bash
 mkdir celestia-client-example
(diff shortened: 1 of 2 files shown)


gbarros:
So, we are blocked until we can get tokens approved for the workflow

```
Error: The deploy step encountered an error: remote: Permission to celestiaorg/docs-preview.git denied to jcstein.
fatal: unable to access 'https://github.com/celestiaorg/docs-preview.git/': The requested URL returned error: 403 ❌
Notice: Deployment failed! ❌
```

gbarros:
I am closing this as it's outdated and I've already created something similar.

PR #2313, opened 2025-11-19
docs: add learn tia section

jcstein:

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -43,5 +43,5 @@ jobs:
         uses: lycheeverse/lychee-action@v2.0.0
         with:
           fail: true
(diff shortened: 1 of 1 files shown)


app/learn/TIA/_meta.js
```diff
@@ -1,7 +1,7 @@
 const meta = {
-  "overview": "Overview",
-  "governance": "Governance",
-  "defi": "DeFi",
+  "staking-governance-supply": "Staking, governance, & supply",
+  staking: "Staking on Celestia",
```
gbarros, line 4: The menu is looking clean.
I wonder if we could fix just the mechanics (but that's a different PR)

jcstein:
marking as draft so i dont forget to change filetypes

PR #2331, opened 2025-12-09
Old docs redirects

gbarros:
This pull request introduces support for custom redirects in the Next.js configuration by reading from a new `redirects.json` file. It also makes minor adjustments to the Next.js export options. The most important changes are grouped below:

**Redirects feature:**

* Added an asynchronous `redirects()` function to `next.config.mjs` that loads redirect rules from the new `redirects.json` file, enabling centralized management of redirects.
* Created the `redirects.json` file, which contains a list of permanent redirect mappings for various routes.

**Configuration updates:**

* Removed the `output: 'export'` option from the Next.js configuration, possibly to improve compatibility with the new redirects setup.
* Added imports for Node.js `fs` and `path` modules in `next.config.mjs` to support reading the redirects file.


jcstein:
i wonder...if there's a way we can export this to work with static site 🤔

redirects.json
```diff
@@ -0,0 +1,47 @@
+[
```
jcstein, line 1: my only thoughts here are that maybe we should add an exhaustive list if this works smoothly. however, i can't see them 😅 
Screenshot 2025-12-09 at 11 46 36

jcstein:
does this commit add ability for this to work on deployed gh pages? https://github.com/celestiaorg/docs/pull/2331/files/7c0e5a29084e8329bedca12a58f6aa6812ff4199..f3be8068cc719a6916dd14b8625a03ea3ffec4b4

PR #2337, opened 2025-12-10
Fix broken links

gbarros:
This pull request updates the project's linting and link-checking workflows to use npm instead of yarn, introduces a new npm script for link checking, and updates dependencies to support the new workflow. It also corrects a documentation link.

**Workflow and Tooling Updates:**

* Switched the GitHub Actions workflow in `.github/workflows/lint.yaml` from yarn to npm for dependency installation, caching, and running scripts. The lint and link-check jobs now use `npm install` and `npm run lint`. The link-check job has also been reworked to use a custom npm script instead of the previous third-party action.
* Added a new `check-links` npm script in `package.json` for running the custom link checker, and updated the workflow to use it.
* Updated `package.json` dependencies to include `remark-mdx`, `remark-parse`, `unified`, and `unist-util-visit` to support the new link checking script.

**Documentation:**

* Fixed a broken documentation link in `app/operate/networks/mainnet-beta/page.mdx` to point to the correct path for information about light vs archival nodes.

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -49,4 +49,4 @@ jobs:
         run: npm install --no-package-lock --legacy-peer-deps
 
       - name: Run Link Checker
-        run: npm run check-links -- --all --concurrency=2
+        run: npm run check-links -- --all --concurrency=2 --skip="celestia.org,nexus.hyperlane.xyz"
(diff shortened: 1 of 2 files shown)


jcstein:
should we strictly just use npm on the entire repo to avoid issues between `yarn.lock` and `package-lock.json`?

PR #2396, opened 2026-01-30
fix: correct pbs-proxy to pda-proxy in private blockspace quickstart

jcstein:
## Overview

The correct image name is pda-proxy, not pbs-proxy. Updated all references to use ghcr.io/celestiaorg/pda-proxy:v0.3.0


jcstein:
closing in favor of #2397

PR #2432, opened 2026-02-25
docs: align Foundation Delegation Program commission rule with CIP-44

jcstein:
## Summary
- Replace the hardcoded `25% or less commission` requirement in the Foundation Delegation Program docs.
- Align the requirement with protocol `staking.MaxCommissionRate`.
- Note the transition from 25% to 60% after CIP-44 activation.

## Why
CIP-44 proposes increasing validator max commission from 25% to 60%, so the docs should not hardcode a stale cap.


jcstein:
slop

PR #2485, opened 2026-04-16
fix: resolve react-hooks/set-state-in-effect lint errors

### LINKED ISSUE #2484: Lint: eslint fails on NodeAPIContent.tsx with react-hooks/set-state-in-effect
## Problem

The eslint CI job is failing on PRs with errors in `app/build/rpc/components/NodeAPIContent.tsx`:

- Line 147: `setSelectedVersion(targetVersion)` — setState called synchronously inside a `useEffect`
- Line 160: `fetchJsonData(selectedVersion, false)` — effect calls a function that synchronously sets state

Both trigger the new `react-hooks/set-state-in-effect` rule.

## Why PRs fail but main passes

`.github/workflows/lint.yaml` installs dependencies with `npm install --no-package-lock --legacy-peer-deps`, so every CI run resolves latest versions from npm. A recent minor bump to `eslint-plugin-react-hooks` introduced the `set-state-in-effect` rule; main's last lint run predates the bump, new PRs hit it. Any re-run on main would also fail.

## Repro

Run `npm install --no-package-lock --legacy-peer-deps && npm run lint` on current `main`.

## Proposed fix

1. Derive the initial `selectedVersion` from the URL via a lazy `useState` initializer (removes the setState in the mount effect).
2. Move the version-change fetch into the dropdown change handler instead of a `useEffect` that watches `selectedVersion`.

Optionally: pin `eslint-plugin-react-hooks` in `package.json` or drop `--no-package-lock` from the workflow to prevent silent upstream breakage.

## Failing run

https://github.com/celestiaorg/docs/actions/runs/24539080913/job/71740793049

rootulp:
Closes #2484

## Summary
- Move the version-change fetch into the dropdown `onChange` handler and drop the effect that re-fetched on `selectedVersion` changes — the `fetchJsonData` call is no longer triggered from a dependency-tracked effect.
- Keep the mount-only URL-param read inside `useEffect` so server-rendered output matches client hydration (addresses [Devin's SSR mismatch feedback](https://github.com/celestiaorg/docs/pull/2485#discussion_r3096980103)).
- Suppress `react-hooks/set-state-in-effect` only on the `setSelectedVersion` line used to sync state with the URL after mount, with a comment.

## Test plan
- [x] `npm install --no-package-lock --legacy-peer-deps && npm run lint` passes locally (matches CI)
- [x] CI `eslint` job is green
- [x] Manually verify the RPC docs page: initial load picks up `?version=` from the URL, dropdown switches versions and fetches the correct spec, URL updates on change, no hydration warnings in the console.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/app/build/rpc/components/NodeAPIContent.tsx
+++ b/app/build/rpc/components/NodeAPIContent.tsx
@@ -107,7 +107,11 @@ export default function RPCDocumentation() {
     active: false,
   });
 
-  const [selectedVersion, setSelectedVersion] = useState(versions[0]);
+  const [selectedVersion, setSelectedVersion] = useState(() => {
+    if (typeof window === 'undefined') return versions[0];
+    const versionParam = new URLSearchParams(window.location.search).get('version');
+    return versionParam && versions.includes(versionParam) ? versionParam : versions[0];
+  });
   const [searchTerm, setSearchTerm] = useState('');
 
   // Consolidated fetch function with loading and error handling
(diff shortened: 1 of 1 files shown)


rootulp:
I manually tested and the version drop down still appears to work

Screenshot 2026-04-16 at 4 49 22 PM
Screenshot 2026-04-16 at 4 49 14 PM

jcstein (approved):
Thank you @rootulp, :shipit:

PR #2531, opened 2026-06-18
Update page.mdx

mindstyle85:

### DIFF
--- a/app/operate/consensus-validators/consensus-node/page.mdx
+++ b/app/operate/consensus-validators/consensus-node/page.mdx
@@ -157,11 +157,10 @@ nodes.
 
 #### Retain all block data
 
-And in your `app.toml`, `min-retain-blocks` should remain as the default
-setting of `3000`:
+In your `app.toml`, explicitly set `min-retain-blocks` to `0`:
 
 ```toml
-min-retain-blocks = 3000 # retain 3000 block data, this is default setting
+min-retain-blocks = 0 # retain all block data
 ```
 
 ### Query transactions by hash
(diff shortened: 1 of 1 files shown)


mindstyle85:
we have to update this, im not sure if its anywhere else in the docs, or if we want to have a different setting there, but we at least have to fix the fact that default is no longer 0

app/operate/consensus-validators/consensus-node/page.mdx
```diff
@@ -158,10 +158,10 @@ nodes.
 #### Retain all block data
```
rach-id, line 158: I don't think we should set this to 3000 here. We can say the default is 3000. But to retain all block data, we need to keep `0`

rach-id (commented):
Thanks for updating this

app/operate/consensus-validators/consensus-node/page.mdx
mindstyle85: yeah i know, i guess the question here is what will we recommend

rach-id: we can recommend 600k blocks, however, it shouldn't be under the `Retain all block data` section because that one is specific for archival nodes

mindstyle85: i guess this here are recommended settings, so we should add the setting with 0 somehwere else in the docs, wdyt @jcstein ?

jcstein: I think we should split the guidance here.

For the “Retain all block data” section, we should keep:

`min-retain-blocks = 0`

because this section is specifically for bridge/archive behavior. `3000` is the current default/minimum for normal operation, but it does not retain all block data.

If we want to recommend `600000`, let’s add that as a separate non-archival/pruned-node recommendation under storage/pruning, not under “Retain all block data.” We should also update any wording that says `0` is still the default and instead say to set it explicitly when full block retention is required.

jcstein: https://github.com/celestiaorg/docs/pull/2531/commits/3077f5dd68cb3e1aa9c2148bc0bf18e61bc6fce0

mindstyle85: 600000 should be what we suggest for validators

jcstein: will adjust

mindstyle85: lgtm!

jcstein: cool! should be good in https://github.com/celestiaorg/docs/pull/2531/commits/0749a9201d5ac7d0579f38f1ef80912b18131404

jcstein (approved):

PR #2548, opened 2026-07-30
Fix flaky link checks

jcstein:
## What changed

- remove the retired ZKsync Gateway staging explorer hyperlink while preserving the deployed contract address
- apply the existing 404-only policy before returning non-retryable external-link failures
- apply that policy to Celestia status pages and Mammoblocks, which reproduced transient fetch/502 failures locally

## Root cause

The latest `main` link-check runs consistently reported one stale ZKsync explorer URL and two Etherscan 403 responses. Etherscan address links were already configured to fail only on 404, but the checker returned non-retryable 403 responses before reaching that policy. The same ordering made other configured 404-only hosts vulnerable to false CI failures.

This keeps definitive 404s actionable while treating bot blocking, temporary service failures, and network errors as skipped for the explicitly configured hosts.

## Validation

- `npm run lint`
- `npm run check-links -- --all --concurrency=2 --skip="celestia.org,nexus.hyperlane.xyz"` — 0 broken internal links, 0 broken external links
- `git diff --check`


gbarros (approved):
This is good!

PR #2553, opened 2026-08-04
docs: add OpenChainBench live RPC latency benchmark link

Flotapponnier:
Adds a reference to [OpenChainBench](https://openchainbench.com/benchmarks/celestia-rpc), which independently monitors public Celestia RPC endpoint latency (Tendermint status, p50/p90/p99, 3 probe regions, updated every 60 seconds). Providers tracked: PublicNode, Polkachu, LavenderFive.


Flotapponnier (commented):
Fixed in a761db9 — replaced "these endpoints" with the explicit provider list (PublicNode, Polkachu, LavenderFive) so the note accurately describes what the benchmark page tracks, not the community table endpoints.

Flotapponnier:
Fixed in a761db9 — replaced "these endpoints" with the explicit provider list (PublicNode, Polkachu, LavenderFive) so the note accurately describes what the benchmark page tracks, not the endpoints in the community table above. Thanks for catching this.

jcstein (approved):

PR #2555, opened 2026-08-07
Update celestia-node RPC reference through v0.31.4

jcstein:
## Summary

- add the missing mainnet OpenRPC snapshots for celestia-node v0.30.2, v0.31.3, and v0.31.4, with v0.31.4 as the default reference
- remove the stale fraud namespace from the current Node API overview and correct `node.Info.type` to `string` in the v0.31.x schemas
- document the v0.31.3 `config-update` requirement and 16 MiB RPC request limit
- refresh the LLM resource links and public agent skill to target v0.31.4

## Why

The published Node API reference was pinned to v0.28.4 even though the documented Mainnet Beta node version is v0.31.4. This left removed fraud RPC methods visible and omitted current response-shape and server-limit behavior.

The upstream v0.31.3 and v0.31.4 `openrpc.json` release assets also contain a docgen warning before the JSON document and describe `node.Info.type` as an integer despite the released type's JSON marshaler returning strings. These snapshots remove the warning and correct that schema field while otherwise preserving the release output.

## Validation

- parsed all three new specifications with `jq`
- confirmed v0.30.2 matches the official release asset after JSON normalization
- confirmed the only semantic change from the sanitized v0.31.3 and v0.31.4 assets is `node.Info.type`: `integer` to `string`
- `yarn lint` — passes with one pre-existing warning
- `yarn build` — passes, including static generation, LLM output generation, and Pagefind indexing
- verified the generated site copies all three specs byte-for-byte and includes the new operator guidance
- verified the new v0.31.3 release link returns HTTP 200

`yarn check-links` remains blocked locally by the repository's existing Node 26 `glob`/`minimatch` ESM mismatch (`GLOBSTAR` is not exported).


jcstein (commented):
lgtm, will make issues re: issue on celestia-node with the openrpcgen

gbarros (approved):

PR #2559, opened 2026-08-13
docs: stop recommending the test keyring for Mainnet Beta validators

### LINKED ISSUE #1968: docs: recommend hardware wallet or multisig for production validator, not test keyring
From [Slack message](https://celestia-team.slack.com/archives/C08GGQFHM1S/p1743686792229149?thread_ts=1743439910.224859&cid=C08GGQFHM1S):
> the docs should be corrected and made consistent with what I wrote.
> 1. The page to create a validator wallet should use --ledger and discuss --index, not the test backend. There's not much point for the only docs on running a validator not tell me how to run a production validator.
> 1. The page to create a validator wallet and the page to create a validator should use the same validator key monikey, not both validator and $VALIDATOR_WALLET.

durenodes:
Closes #1968

The wallet guide opened by writing `keyring-backend test` into the CLI config, which makes every later command use it silently, and the validator guide passed `--keyring-backend=test` in all three of its transaction examples.

The Cosmos SDK [documents that backend](https://docs.cosmos.network/sdk/v0.53/user/run-node/keyring) as "a password-less variation of the file backend. Keys are stored unencrypted on disk" and "not recommended for use in production environments". These are the only docs for running a Celestia validator, so following them as written leaves a Mainnet Beta signing key readable by anyone with access to the directory.

## Changes

- Compare the keyring backends and say where each is appropriate, with a warning that points at hardware wallets and the existing [multisig page](https://docs.celestia.org/operate/keys-wallets/multisig).
- Drop the global `config keyring-backend test` line and pass the backend per command, so a testnet choice cannot silently carry into mainnet work.
- The validator guide already had `Mainnet Beta` and `Mocha` tabs carrying the same backend. Mainnet Beta now reads `$KEYRING_BACKEND`; Mocha keeps `test`, which is what the tabs were there to distinguish.
- Both pages use `$VALIDATOR_WALLET`, so the wallet you create is the wallet the next guide expects. This was the second point in #1968.

## Left out on purpose

I did not add step-by-step `--ledger` instructions. I have not set up a Celestia validator with a hardware wallet, and I would rather not document a flow I have not run. The backend table names `ledger` and the warning recommends it. Happy to add the full flow if a maintainer who has done it wants to review the wording.

## Testing

`npm run lint` passes. `npm run build` fails locally on `/_not-found` and two unrelated pages, but it fails the same way on a clean `main` here, so it looks like a local Node version issue rather than something in this change.

---

<a href="https://app.devin.ai/review/celestiaorg/docs/pull/2559" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/app/operate/consensus-validators/cli-reference/page.mdx
+++ b/app/operate/consensus-validators/cli-reference/page.mdx
@@ -39,12 +39,17 @@ Available Commands:
 ## Creating a wallet
 
 ```sh
-celestia-appd config keyring-backend test
+celestia-appd config keyring-backend <backend>
 ```
 
 `keyring-backend` configures the keyring's backend, where the keys are stored.
 
 Options are: `os|file|kwallet|pass|test|memory`.
+
+`test` keeps the key unencrypted on disk with no password, so reserve it for
+local development and testnets. See
+[create a wallet with celestia-app](/operate/keys-wallets/celestia-app-wallet)
+for a comparison of the backends.
 
 You can learn more on the [Cosmos documentation](https://docs.cosmos.network/sdk/v0.53/user/run-node/keyring)
 or [Go Package documentation](https://pkg.go.dev/github.com/cosmos/cosmos-sdk/crypto/keyring).
--- a/app/operate/consensus-validators/validator-node/page.mdx
+++ b/app/operate/consensus-validators/validator-node/page.mdx
@@ -26,6 +26,9 @@ First, follow the instructions on
 ### Wallet
 
 Follow [the tutorial on creating a wallet](/operate/keys-wallets/celestia-app-wallet).
+That page also covers the available keyring backends. Pick one before you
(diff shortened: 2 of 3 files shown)


durenodes:
Both findings were correct — fixed in 6fb2e6d.

The wallet guide now passes `--keyring-backend` on every command that touches a key, so `keys list` and the management examples no longer look in a different keyring than the one `keys add` just wrote to. The Mocha tab uses `$KEYRING_BACKEND` instead of hardcoding `test`, and the export step says where `test` is acceptable rather than a tab quietly overriding it. `keys show --bech val` and the delegate example were missing the flag for the same reason.

I also updated `cli-reference`, which still opened its wallet section with `celestia-appd config keyring-backend test` — the line this PR removes elsewhere.

